Issue #21, opened 2024-12-04
Predicted completion times are calculated incorrectly


Razenpok:
### Description

https://discord.com/channels/667551433503014924/1311931276684689408

PR #48, opened 2025-02-15
Fix Production line queue time is calculated incorrectly (2nd try)

### LINKED ISSUE #43: Production line queue time is calculated incorrectly
### Description

The time that appears below the production orders that shows when it will be completed is calculated wrong if you have more than one type of recipe going in your production line. I have 7 HER recipes that all complete before 2 GRN recipes. If I add a GRN or BEA or HCP recipe to queue, it calculates the time of completion from the first GRN recipe instead of the first recipe to complete in the whole production line, which would be an HER recipe.

If you only have one recipe taking up all the slots in the production line, I'm fairly certain it calculates the queue times correctly.

Example: Notice the first picture, GRN recipe ends in 2:15pm+1d. The queued recipe is 5hr 6m long. its ending time is added onto this GRN recipe instead of the HER recipe that completes in about 3 hr. I included 3 pictures with 3 short recipes to show that they calculate from the first-to-complete GRN recipe instead of the HER recipe that completes in about 3 hrs.

<details>

Image
Image
Image
</details>

1. Have 9 FRM
2. Have 7 HER recipes in queue that all finish before 2 GRN recipes
3. Any queued recipe will calculate its finish time based upon the first-to-complete GRN recipe

### Extension version

25.2.6.1805

### Browser(s) used

Firefox

CoroNaut:
Fixes #43

The issue is line 69:
queue.shift()! is getting the first completion timestamp inside the queue variable. This array is not sorted properly to be able to get the recipe that will complete _**first**_ and is instead getting timestamps closer to the end of the currently producing recipes.
https://github.com/refined-prun/refined-prun/blob/bf18dcba9f0178808732a98da02ca6ee9c29957c/src/features/basic/prod-order-eta.ts#L57-L74

Adding `queue.sort()` before line 69 `queue.push(queue.shift()! + lineOrder.duration.millis); ` fixes the issue.
This ensures that the value returned from queue.shift()! is the first recipe to complete. In addition, because the `queue` adds queued recipes, it ensures that all subsequent queued recipes completion times are calculated correctly.

Picture 1 just to show completion times of the first few recipes currently producing.
Picture 2 shows proper completion times for BEA recipes after `queue.sort()` was put in:

<details>

Image

Image
</details>

(First PR, I might make mistakes with it lol)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Fixed
 
 - Fix duplication of Materials in Transit asset value in Long-Term Materials Receivable
+- `prod-order-eta`: Fix completion time being calculated incorrectly
 
 ## 25.2.11
 
--- a/src/features/basic/prod-order-eta.ts
+++ b/src/features/basic/prod-order-eta.ts
@@ -66,6 +66,7 @@ function calcCompletionDate(line: PrunApi.ProductionLine, order: PrunApi.Product
       queue.push(Date.now() + lineOrder.duration.millis);
     } else {
       // Order has not started
+      queue.sort()
       queue.push(queue.shift()! + lineOrder.duration.millis);
     }
     if (lineOrder === order) {


Razenpok:
Thank you for such a detailed PR 🔥 

This was also reported in #21, so this PR will close two issues at once.

Add a changelog entry in Fixed:
> - `prod-order-eta`: Fix queue time being calculated incorrectly

CoroNaut:
I put the fix into a separate branch and added the changelog summary.

PR #54, opened 2025-02-22
cxos-centeralign-delete-filled

CoroNaut:
https://com.prosperousuniverse.com/t/move-cxos-delete-filled-to-the-left/6303
Someone on the forum had an issue with the delete button being close to the commodity search bar in CXOS. It may be a rather old request for this feature, but it was simple enough to make. This PR moves the 'DELETE FILLED' button to the left.

The specificity in applyCssRue is required because there are a few other components in the game with similar structure like the CX commodity dropdown box. This ensures that only the delete button is targeted for the applied css rule.

Pictures before/after:

<details>

Capture
Capture2

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,9 @@
 ### Added
 
 - `XIT SHEET`: Add an optional parameter for Sheet ID
-- `cxos-leftalign-delete-filled`: New feature. Moves 'Delete Filled' button to the left in CXOS
 - `context-controls-no-hover`: New feature. Prevents the context controls from displaying description while hovering
   over
+- `cxos-leftalign-delete-filled`: New feature. Moves 'Delete Filled' button to the left in CXOS
 - `prod-hide-percent`: New feature. Hides percent value from production lines.
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


Razenpok:
I don't think this is a good idea because this button is now close to the selected material filters (and they are even smaller than the search box, so there's a bigger chance for misclick)
image

CoroNaut:
How about centering it instead? "justify-content: center;" ?

Capture3

Razenpok:
tbh this one looks out of place compared to other PrUn UI. What about moving the search bar to the left instead?

CoroNaut:
I agree that moving the search bar left would look good. I also agree the delete button looks weird in the center, but its the best I got as a free player. I've considered moving the delete button below the list entirely, but that requires watching the html for updates to check when a CXOS buffer appears on screen, which would be a bit jank for a simple feature.

Unfortunately I don't have PRO and I simply don't get the HTML for all of those features. I don't know the class for it or what the structure looks like to be able to test it. I'm thinking it would probably be against ToS for me to ask to get that html from someone too. I'll close this PR and make an issue for it instead, perhaps in the future I can implement it.

PR #56, opened 2025-02-23
Feature to compress the weight/volume information in an inventory with id parameter

CoroNaut:
The 3 rows with planet, weight, volume bulks up the inventory window no matter what size it is. This PR adds a feature to compress this information into a bar on top. The planet link becomes a standard context item in the top-right while the weight and volume progress bars are neatly placed to the left. This also interacts nicely with the feature "hide-weight-volume-labels".

A new container is made to house this information and the original 100% wide row is hidden.

Pictures:
1. No feature
2. Feature "inv-compress-inventory-info" on
3. Feature "hide-weight-volume-labels" on with compression
4. Smaller window size

<details>

Capture1

Capture2

Capture3

Capture4

</details>


Razenpok:
Alright, there's a lot to unpack here:

1. If you have more files than a pair of .ts + .module.css files, you need to create a folder for a feature. Take a look at any other feature that is inside its own folder.

2. Always name vue components the same as the name you will import. Since you `import ContextRow from './inv-compress-inventory-info.vue'`, the file name must be `ContextRow.vue`, not `inv-compress-inventory-info.vue`

3. There's an easier way to compress the inventory header:
  i. Set `display: none;` in `.StoreView__column___x_hGAq3 .StoreView__capacity___kzfUS7E:nth-child(1)`
  ii. Set `flex-direction: row` in `StoreView__column___x_hGAq3`

4. To add a context bar with buttons, use the `ContextControls` component from the `components` folder. This component is the 1-to-1 copy of the in-game context bar. Be sure to put it in the right place in the node hierarchy, so it is in the exact same spot as any other context bar.

image

5. To put the weight/volume row on the same line as context bar, add `position: absolute; top: -16px; left: -2px;` to `StoreView__row___PEJZ6k0`

6. You can add `padding-top: 0; padding-bottom: 0;` to `InventorySortControls__controls___qk5heAZ` for an even more compressed look.

7. As planets can have inventories, stations can too, so only showing `PLI` in the context bar won't work. For example, a CX warehouse at ANT must show `STNS ANT` instead. So, you need to figure out if it's a planet inventory or a station warehouse. Which leads us to a big topic of...

## How do I identify a thing in the UI?

Unless there's no other way, never rely on strings in HTML (yes, I am looking at you, `extractPlanetId`). This is a very verbose way that is also very unstable and can break when prun devs change something in the UI. And handling different localizations is a pain in the ass.

Instead, use ids or similar stable ways to identify a thing. For example, in `INV e0a8b4e1` we have `e0a8b4e1` - that's an inventory id with which we can find the inventory and, ultimately, the address. This approach won't suddenly break unless there will be a big change in the underlying API data, but then we are screwed anyway. 

Speaking of `INV`, this shit is extremely annoying to deal with, because it can accept both short and long ids, as well as the planet/station id (called natural id). So, if I have an inventory with an id `e0a8b4e137214edf0e044c44` on the planet `ZV-307b`, **all of these commands will open the same inventory**:
- `INV e0a8b4e1`
- `INV e0a8b4e137214edf0e044c44`
- `INV e0` (if I have no other inventories with an id starting with `e0`)
- `INV ZV-307b` (if I have only a single BS or WAR inventory on the planet, otherwise it will open a list of inventories)

Look how different it looks like for `INV ZV-307b` and `INV RC-040b` in my account:
image

So, you have a `tile.parameter` with either a store id or a natural id; how to get a natural id and figure out if we need to use `PLI` or `STNS` from that?

To do that, use the `getInvStore`. You can look how it is used in the `custom-item-sorting` feature. It accepts the `INV` parameter and returns the first store it finds that matches the parameter. The store is of type `PrunApi.Store` (you can find it in the project). It has the `type` field, and we only need two types here:

(Note: `store` is a variable that was returned by `getInvStore`. Other things, like `sitesStore` are imported from other files. Unfortunately, there is an overlap in terms: "store" as in "storage/inventory" and "store" as in "Vue reactive store")

1. If it's a `STORE`, you will need to find the site using `sitesStore.getById(store.addressableId)`. Then, get a natural id using `getEntityNaturalIdFromAddress(site.address)`. You can use this natural id with `PLI` in the context bar.
2. If it's a `WAREHOUSE_STORE`, you will need to find the warehouse using `warehousesStore.getByStoreId(store.id)` (see below for implementation). Then, get a natural id using `getEntityNaturalIdFromAddress(warehouse.address)`. Then, get a location line using `getLocationLineFromAddress(warehouse.address)`. If `location?.type === 'STATION'`, then use `STNS` with the natural id. If `location?.type === 'PLANET'`, then use `PLI` with the natural id.

In cases you get `undefined` at some point of doing all that, just add a plain `INV` command to the context bar.

Here's the implementation of `warehousesStore.getByStoreId`:

```ts
// warehouses.ts
// ...

const getByStoreId = createMapGetter(state.all, x => x.storeId);

export const warehousesStore = {
  ...state,
  getByEntityNaturalId,
  getByStoreId,
};
```

---

Here's the end result (manually changed the DOM in the dev tools):
image
Notice how this approach requires no custom Vue components.

CoroNaut:
As you suggest, much simpler code and implementation and very good ideas.
A few pictures of the redesign:

<details>

Capture1
Capture2
Capture3
Capture4


</details>

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  createFragmentApp(ContextControls, { items }).before(tile.anchor.parentElement as HTMLDivElement);
+}
+
+function init() {
+  applyCssRule(`.${C.StoreView.row}`, classes.reposition);
+  applyCssRule(`.${C.StoreView.column}`, classes.flexRow);
+  applyCssRule(`.${C.StoreView.column} .${C.StoreView.capacity}:nth-child(1)`, css.hidden);
+  applyCssRule(`.${C.InventorySortControls.controls}`, classes.padLeftRight);
```
Razenpok, line 64: Two things here:
1. If you apply a css rule to a single class, use `applyClassCssRule`, so:
```ts
  applyClassCssRule(C.StoreView.row, classes.reposition);
```
2. If a feature is applied to specific commands (`INV` in this case), use the `applyScoped` functions, so:
```ts
  applyScopedClassCssRule('INV', C.StoreView.row, classes.reposition);
```
Because, for example, as it is right now, this feature will leak these css overrides to the `SHPI` command.

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,68 @@
...
+    case 'FTL_FUEL_STORE':
+    case 'STL_FUEL_STORE': {
+      const shipStore = shipsStore.getById(store?.addressableId);
+      const shipRegistration = shipStore?.registration;
+      items[0].cmd = `SHP ${shipRegistration}`;
+      break;
+    }
+  }
+
+  createFragmentApp(ContextControls, { items }).before(tile.anchor.parentElement as HTMLDivElement);
```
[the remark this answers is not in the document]
CoroNaut: `tile.anchor` returns an `HTMLDivElement`
`tile.anchor.parentElement` returns `HTMLElement | null`

The casting is there to ignore `null` since the `tile.anchor` will always have a parent element, and theres no async problems or otherwise that I can see that will make `tile.anchor.parentElement` null.

Alternatively, I can put the `createFragmentApp` inside 
`if (tile.anchor.parentElement) { createFragmentApp... }`
which also ensures null is ignored.

CoroNaut:
I'm opposed to applying this feature to `SHPI`. It already has context controls and has the unload button which will make things complicated if needing to account for those issues in this specific feature.

src/features/advanced/inv-compress-inventory-info.ts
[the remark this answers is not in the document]
Razenpok: Ah, then you can write `tile.anchor.parentElement!` instead of `tile.anchor.parentElement as HTMLDivElement`. The exclamation mark tells the compiler to ignore `null` and `undefined`

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,77 @@
...
+      break;
+    }
+  }
+  createFragmentApp(ContextControls, { items: [{ cmd: cmd ?? 'INV' }] }).before(
+    tile.anchor.parentElement as HTMLDivElement,
+  );
+}
+
+function init() {
+  applyScopedClassCssRule('INV', `${C.StoreView.row}`, classes.storeInfo);
```
Razenpok, line 66:
```suggestion
  applyScopedClassCssRule('INV', C.StoreView.row, classes.storeInfo);
```

Don't wrap a single variable in `${}` (same in other `applyScopedClassCssRule`)

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,77 @@
...
+import { sitesStore } from '@src/infrastructure/prun-api/data/sites';
+import {
+  getEntityNaturalIdFromAddress,
+  getLocationLineFromAddress,
+} from '@src/infrastructure/prun-api/data/addresses';
+import { warehousesStore } from '@src/infrastructure/prun-api/data/warehouses';
+import { shipsStore } from '@src/infrastructure/prun-api/data/ships';
+
+async function onTileReady(tile: PrunTile) {
+  if (!tile.parameter || tile.command === 'SHPI') {
```
Razenpok, line 15:
```suggestion
  if (!tile.parameter) {
```

You don't need to check for `SHPI` since you only observe `INV` tiles:
```ts
tiles.observe('INV', onTileReady);
```

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const store = getInvStore(tile.parameter);
+  if (!store) {
+    return;
+  }
+
+  let cmd = '';
+  switch (store.type) {
+    case 'STORE': {
+      const siteStore = sitesStore.getById(store?.addressableId);
```
Razenpok, line 28:
```suggestion
      const site = sitesStore.getById(store?.addressableId);
```

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,77 @@
...
+      const siteEntityNaturalId = getEntityNaturalIdFromAddress(siteStore?.address);
+      if (siteEntityNaturalId) {
+        cmd = `PLI ${siteEntityNaturalId}`;
+      }
+      break;
+    }
+    case 'WAREHOUSE_STORE': {
+      const warehouseStore = warehousesStore.getById(store?.addressableId);
```
Razenpok, line 36:
```suggestion
      const warehouse = warehousesStore.getById(store?.addressableId);
```

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,77 @@
...
+        cmd = `PLI ${warehouseEntityNaturalId}`;
+      } else if (location?.type === 'STATION') {
+        cmd = `STNS ${warehouseEntityNaturalId}`;
+      }
+      break;
+    }
+    case 'SHIP_STORE':
+    case 'FTL_FUEL_STORE':
+    case 'STL_FUEL_STORE': {
+      const shipStore = shipsStore.getById(store?.addressableId);
```
Razenpok, line 52:
```suggestion
      const ship = shipsStore.getById(store?.addressableId);
```

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,77 @@
...
+  const store = getInvStore(tile.parameter);
+  if (!store) {
+    return;
+  }
+
+  let cmd = '';
+  switch (store.type) {
+    case 'STORE': {
+      const siteStore = sitesStore.getById(store?.addressableId);
+      const siteEntityNaturalId = getEntityNaturalIdFromAddress(siteStore?.address);
```
Razenpok, line 29:
```suggestion
      const naturalId = getEntityNaturalIdFromAddress(site?.address);
```

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,77 @@
...
+      if (siteEntityNaturalId) {
+        cmd = `PLI ${siteEntityNaturalId}`;
+      }
+      break;
+    }
+    case 'WAREHOUSE_STORE': {
+      const warehouseStore = warehousesStore.getById(store?.addressableId);
+      const warehouseEntityNaturalId = getEntityNaturalIdFromAddress(warehouseStore?.address);
```
Razenpok, line 37:
```suggestion
      const naturalId = getEntityNaturalIdFromAddress(warehouse?.address);
```

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,77 @@
...
+      } else if (location?.type === 'STATION') {
+        cmd = `STNS ${warehouseEntityNaturalId}`;
+      }
+      break;
+    }
+    case 'SHIP_STORE':
+    case 'FTL_FUEL_STORE':
+    case 'STL_FUEL_STORE': {
+      const shipStore = shipsStore.getById(store?.addressableId);
+      const shipRegistration = shipStore?.registration;
```
Razenpok, line 53:
```suggestion
      const registration = ship?.registration;
```

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+function init() {
+  applyScopedClassCssRule('INV', `${C.StoreView.row}`, classes.storeInfo);
+  applyScopedClassCssRule('INV', `${C.StoreView.column}`, classes.storeInfoColumn);
+  applyScopedClassCssRule(
+    'INV',
+    `${C.StoreView.column} .${C.StoreView.capacity}:nth-child(1)`,
+    css.hidden,
+  );
```
Razenpok, line 72:
```suggestion
  applyScopedCssRule(
    'INV',
    `.${C.StoreView.column} .${C.StoreView.capacity}:nth-child(1)`,
    css.hidden,
  );
```

This is a complex rule, use `applyScopedCssRule` instead

Razenpok (changes requested):

src/features/advanced/inv-compress-inventory-info.module.css
```diff
@@ -0,0 +1,14 @@
+.sortControls {
+  padding-top: 0px;
```
Razenpok, line 2:
```suggestion
  padding-top: 0;
```
Minor improvement: in css, you can amend "px" in "0px"

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,80 @@
...
+    return;
+  }
+  await $(tile.anchor, C.StoreView.container);
+
+  const store = getInvStore(tile.parameter);
+  if (!store) {
+    return;
+  }
+
+  let cmd = '';
```
Razenpok, line 28:
```suggestion
  let cmd: string | undefined;
```
Otherwise, `cmd ?? 'INV'` won't work because `''` is not a nullish value (`??` only checks for `null` and `undefined`)

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,80 @@
...
+import { shipsStore } from '@src/infrastructure/prun-api/data/ships';
+
+async function onTileReady(tile: PrunTile) {
+  if (!tile.parameter) {
+    return;
+  }
+  await $(tile.anchor, C.StoreView.container);
+
+  const store = getInvStore(tile.parameter);
+  if (!store) {
```
Razenpok, line 24: I don't see why we need this `if`. Every place we access the store, we do it via `store?.`. And if, for some reason, no store was found, we still probably want to display `ContextControls` with `INV`. Otherwise, we risk a broken layout because we apply absolute positioning to`C.StoreView.row`.

One place where you will need to change is `switch (store.type) {` to `switch (store?.type) {`

src/features/advanced/inv-compress-inventory-info.ts
```diff
@@ -0,0 +1,80 @@
...
+import { sitesStore } from '@src/infrastructure/prun-api/data/sites';
+import {
+  getEntityNaturalIdFromAddress,
+  getLocationLineFromAddress,
+} from '@src/infrastructure/prun-api/data/addresses';
+import { warehousesStore } from '@src/infrastructure/prun-api/data/warehouses';
+import { shipsStore } from '@src/infrastructure/prun-api/data/ships';
+
+async function onTileReady(tile: PrunTile) {
+  if (!tile.parameter) {
```
Razenpok, line 18: This `if` was initially added here to prevent running in plain `INV` command without parameters (my guess). But now it's safe to remove this `if` because we await for `C.StoreView.container`. The inventory lists (like in plain `INV`) wont get past this await.

Razenpok (changes requested):
Aight, last batch of changes, and then imma merg

Good job with this PR, the feature is awesome. I'll publish a release after merging it 👍

PR #58, opened 2025-02-26
Quick fix to inv-compress-inventory-info

### LINKED ISSUE #57: requesting adjustments to inv-compress-inventory-info
### Description

- additional space allocated to top and bottom when buffer/tile width is below 400px, or for two lines to be used
- re-add name of location. button for the PLI is great, but with multiple tiles on a tab, it makes it hard to differentiate between different WARs. This could, say, go up in the "TileFrame_title" perhaps?
"INVENTORY WAR" -> "WAR HARMONIA"

CoroNaut:
Fixes #57

This PR gives simple rules to the inv-compress-inventory-info feature and addresses concerns about the missing inventory location link and overlapping text. I'm not sure if you'd be interested in a quick release to alleviate user concerns about the feature.

I'm thinking the context controls could be added in the inventory to a separate feature with many more useful buttons rather than in this feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Fixed
+
+- `inv-compress-inventory-info`: Fix overlapping text and displays inventory location link
+
 ## 25.2.25
 
 ### Added
--- a/src/features/advanced/inv-compress-inventory-info.module.css
+++ b/src/features/advanced/inv-compress-inventory-info.module.css
@@ -1,14 +1,15 @@
-.sortControls {
-  padding-top: 2px;
-  padding-bottom: 2px;
+.progressBarContainer {
+  display: contents;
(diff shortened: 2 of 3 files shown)


CoroNaut:
Changed to make it look like this picture below instead:
Capture

Razenpok:
I find it extra funny that this feature got shrunk into two lines lol
I'll add a couple of changes and make a hotfix release

CoroNaut:
I put the other 99% into another feature that I'll make a PR for it after this is all done. I added a few other buttons to it too.

Razenpok:
I made it so weight and volume labels always go together if there's enough space; and I've returned the removal of padding in `C.InventorySortControls.controls`

https://github.com/user-attachments/assets/147dac45-bf91-4543-af69-26294c45acd3

PR #59, opened 2025-02-27
Feature 'inv-add-context-controls'

CoroNaut:
Adding some context controls to inventories would help someone orient themselves from the current inventory. What other inventories are nearby which can transfer resources? What inventories are in the system? What object is holding that inventory? These commands should help with that. Suggestions are welcome.

This feature will add 4 buttons to `INV` buffers when applicable:
1. `INV`: Simply open all inventories
2. `INV {system}`: gets the system the inventory is in and displays all inventories inside the system
3. `INV {planet/station}`: gets the nearby inventories, inventories you can transfer from/to
4. `PLI/STNS/SHP`: shows the info window for where the inventory is located.

Ships that are on the ground get all the commands, ships that are currently flying only get the 'INV' and 'SHP' commands, I'm not sure how I can get the system from a currently flying ship, if its possible. The `getSystemNaturalIdFromAddress` function returns 'undefined' when flying and a proper system when landed.

I can fix the changelog for this feature if you were going to make a full release for the `inv-compress-inventory-info` first.


Razenpok:
People are very defensive about vertical space, so I'm unsure if people will like it. I've started a poll in Discord - we'll see in a week or two how people react to the proposed change
image

Razenpok:

Razenpok:
The poll results are in, with only 56% of yays. This means that many users will likely disable this feature, hence I'm not going to add it to rprun. I think the bottom line should be at least 75% of yays for adding a controversial/intrusive feature.

image

PR #60, opened 2025-02-27
Fixes incorrect bolding of command arguments in controls

CoroNaut:
`ContextControlsItem.vue` has a slightly different structure from how prosperous universe displays context controls. 
Context controls in HTML from prosperous universe follow this structure:
`<span><span class="ContextControls__cmd___BXQDTL_">SYSI</span> VH-331</span>`

Context controls from the current `ContextControlsItem.vue` are in a different structure:
https://github.com/refined-prun/refined-prun/blob/6395df80e79079bf76eef815ca444f9e696c27ab/src/components/ContextControlsItem.vue#L12-L14

The command as well as the arguments are placed into the `cmd` variable and then into the span with the `C.ContextControls.cmd` which bolds all the text inside. This PR splits the argument from the command and places the argument outside this span.

The tests I did to make sure this worked as intended:
1. cmd='INV' -> command='INV', argument=''
2. cmd='INV 29927187' -> command='INV', argument='29927187'
3. cmd='XIT BURN VH-331a' -> command ='XIT BURN', argument = 'VH-331a'

As such, it basically takes the last string with no spaces and makes that the argument. if there is no argument in cmd, the string goes into command.

I figured modifying this file was easier than going through all the files that make a ContextControlsItem container and changing up all the functions to provide a command and argument separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Fixed
+
+- Fixed incorrect bolding of command arguments in controls
+
 ## 25.2.27
 
 ### Fixed
--- a/src/components/ContextControlsItem.vue
+++ b/src/components/ContextControlsItem.vue
@@ -3,9 +3,9 @@ import { showBuffer } from '@src/infrastructure/prun-ui/buffers';
 
 const props = defineProps<{ cmd: string; label?: string }>();
 
-const lastSpaceIndex = props.cmd.lastIndexOf(' ');
-const command = props.cmd.substring(0, lastSpaceIndex);
-const argument = props.cmd.substring(lastSpaceIndex + 1);
+const index = props.cmd.lastIndexOf(' ');
+const [command, argument] =
+  index > -1 ? [props.cmd.substring(0, index), props.cmd.substring(index + 1)] : [props.cmd, ''];
 
 const itemClasses = [C.ContextControls.item, C.fonts.fontRegular, C.type.typeSmall];
 </script>
--- a/src/components/ContextControlsItem.vue
+++ b/src/components/ContextControlsItem.vue
@@ -1,7 +1,11 @@
 <script setup lang="ts">
 import { showBuffer } from '@src/infrastructure/prun-ui/buffers';
 
-defineProps<{ cmd: string; label?: string }>();
+const props = defineProps<{ cmd: string; label?: string }>();
+
+const lastSpaceIndex = props.cmd.lastIndexOf(' ');
+const command = props.cmd.substring(0, lastSpaceIndex);
+const argument = props.cmd.substring(lastSpaceIndex + 1);
 
 const itemClasses = [C.ContextControls.item, C.fonts.fontRegular, C.type.typeSmall];
 </script>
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Unreleased
+
+### Fixed
+
+- Fixed incorrect bolding of command arguments in context controls
```
Razenpok, line 5:
```suggestion
- Fix incorrect bolding of command arguments in context controls
```

Keep changelog notes in imperative mood

src/components/ContextControlsItem.vue
```diff
@@ -1,7 +1,11 @@
 <script setup lang="ts">
 import { showBuffer } from '@src/infrastructure/prun-ui/buffers';
 
-defineProps<{ cmd: string; label?: string }>();
+const props = defineProps<{ cmd: string; label?: string }>();
+
+const index = props.cmd.lastIndexOf(' ');
+const [command, argument] =
+  index > -1 ? [props.cmd.substring(0, index), props.cmd.substring(index + 1)] : [props.cmd, ''];
```
Razenpok, line 8: Use `computed` if you need to calculate something from props or reactive things

See ShipmentIcon.vue for an example

Razenpok (changes requested):
Instead of making everything bold except the last word in the command, do the opposite: make only the first word bold (or two, if it's a XIT command).

CoroNaut:
I checked "XIT FIN" and "PROD 867ff99d" context controls. they both have a 'XIT' context controls item with an argument that I confirmed as working correctly with my latest commit. I also tested it for regular commands and the MTRA command that has multiple arguments. 

Also when trying to drag-and-drop an item from my base into a ship over the 'AMT' area, it generates an MTRA command that looks weird like:
`MTRA HER                         from-717f5408 to-4ffbecaa`
It generated a lot of needless spaces and I'm not sure if it's a refined prun problem. Anyways, it works correctly with the new computed value for command and arguments.

src/components/ContextControlsItem.vue
```diff
@@ -1,7 +1,14 @@
...
 
-defineProps<{ cmd: string; label?: string }>();
+const props = defineProps<{ cmd: string; label?: string }>();
+
+const commandAndArg = computed(() => {
+  const words = props.cmd.split(' ');
+  return words[0] === 'XIT'
+    ? [words[0] + ' ' + words[1], words.slice(2).join(' ') ?? '']
+    : [words[0], words.slice(1).join(' ') ?? ''];
+});
```
Razenpok, line 11: Let's simplify a bit

```suggestion
const commandAndArg = computed(() => {
  const words = props.cmd.split(' ');
  let command = words.shift();
  if (command === 'XIT') {
    command += ' ' + words.shift();
  }
  return [command, words.join(' ')];
});
```

Also:
1. Rename it to `commandParts` - sounds a bit more natural
2. You don't need the `??` operator here since `words.join(' ')` is never `null` or `undefined`. Read https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

Razenpok (changes requested):

Razenpok:
The weird `MTRA` command you mentioned is generated by the base game, so idk what's going on with it

Razenpok (approved):

PR #61, opened 2025-03-01
Feature 'prod-missing-input-background'

CoroNaut:
In prosperous universe in the PROD, PRODQ, and PRODCO windows, there is little red text saying that the inputs for that recipe are missing. Sometimes I pass by this information or take required inputs out of my base accidently and now the queued recipe can't start producing. This PR replaces the small error text with a much more prominent redish background. I dimmed down this red background from the standard error colors to make it not impose on the user too much. Suggestions welcome

The PRODQ menu has a hover css that adds a transparent white background when hovering over, so the top error there is what it looks like being hovered over. The css is also animated to move the background slightly just for aesthetics. 

Capture


Razenpok:
A better highlighting of orders with errors is a good idea, but this is way too imposing:
1. prun uses this background sparingly and only in the most critical error/danger places. A failing order doesn't fit into this category
2. prun generally has a toned down interface, and this highlight *really* stands out, breaking the overall feel

CoroNaut:
I understand its a bit imposing, I could just make it a simple light red background. I wanted to stay away from coloring text because sometimes that makes it harder to read.

CoroNaut:
Coloring is basically red background at 15% transparency.
Capture

Razenpok:
What about `rgba(217, 83, 79, 0.1);`? It is less bright but still recognizable and blends better with the prun UI gamma since it uses the same color as the red text and the "cancel" button
image

src/features/advanced/prod-missing-input-background.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+function init() {
+  //Error '(intermediate value).SiteProductionLines is undefined' if cmdSpecifics is declared outside of init()
+  cmdSpecifics = {
+    PROD: [3, C.SiteProductionLines.column, C.OrderStatus.error],
+    PRODQ: [2, C.ProductionQueue.table, C.OrderStatus.error],
+    PRODCO: [3, C.InputsOutputsView.inputMaterials, C.InputsOutputsView.amountMissing],
+  };
+  applyScopedClassCssRule(['PROD', 'PRODQ'], C.OrderStatus.error, css.hidden);
```
Razenpok, line 37: Hiding the error text is a very bad idea here, since the error could be anything from "input missing" to "storage full" to "no funds" to something else. The player will likely be unable to guess which one it is correctly, and thus will be forced to hunt down the reason manually.

src/features/advanced/prod-missing-input-background.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (errorContainer) {
+      watchEffectWhileNodeAlive(errorElement, onCleanup => {
+        errorContainer.classList.add(classes.inputMissingContainer);
+        onCleanup(() => {
+          errorContainer.classList.remove(classes.inputMissingContainer);
+        });
+      });
+    }
+  });
+}
```
Razenpok, line 28: You can implement this feature using `:has` selector:
```css
.InputsOutputsView__input___XDlgdLx:has(.InputsOutputsView__amountMissing___kgdxn7l) {
  // this selects InputsOutputsView__input___XDlgdLx element
  // when InputsOutputsView__amountMissing___kgdxn7l
  // is present in some descendant
}
```
With it, you will probably need no supporting js code

src/features/advanced/prod-missing-input-background.ts
```diff
@@ -0,0 +1,51 @@
...
+    `.${classes.inputMissingContainer}:hover > td`,
+    classes.queueDataHover,
+  );
+  tiles.observe(['PROD', 'PRODQ', 'PRODCO'], onTileReady);
+}
+
+features.add(
+  import.meta.url,
+  init,
+  'PROD: Replaces missing input label with red background coloring',
```
Razenpok, line 50: Since this feature touches more than one command, it should not start with "PROD:"

Razenpok (changes requested):

src/features/advanced/prod-missing-input-background.ts
CoroNaut: Very interesting and very powerful css tool, I thought css was limited by having to chain selectors together and you had to do funky stuff. I'll probably be using it in the future, ty for letting me know about that.

CoroNaut:
Updated with the color you suggested and made it all css manipulation.

Razenpok:
I've pushed a couple of changes, fyi:
1. Rewrote order highlighting in PRODQ to work like in table-rows-alternating-colors
2. Since this feature no longer hides any information and just enhances UI, it can be considered a "basic" feature
3. production-order-error-highlight and new description fit the feature better after the changes

PR #63, opened 2025-03-06
Add market-context-menu

CoroNaut:
It usually takes quite a few clicks to get the market information from any particular material. I might pop up the 'CX' and search for it and get this information, but it could be much easier. This PR adds a right-click context menu for any material on screen. Simply right-click the material and you can see market buttons related to that material. With this feature, you can land a ship at the exchange and right-click trade all the items in the ships inventory, very easy! There is also quick access to changing what exchange you get this information from.

I've gotten permission for this feature specifically from the main prosperous universe forum if it were to be in question: [Custom extension feature question: Is this tooltip feature against any rules?](https://com.prosperousuniverse.com/t/custom-extension-feature-question-is-this-tooltip-feature-against-any-rules/6884). Instead of it being a tooltip, its now a simple right-click context menu. 

I am interested in making it possible to have it display as a tooltip through another options as well. Tooltip: hover over material to see context menu. Context menu: right-click material. I've considered adding this setting straight into the context menu as another sub-contextmenu like how hovering over exchange works. Alternatively, I could add a setting in the gameplay tab or adding another tab to 'XIT SET FEAT' to switch the context menu between a contextmenu and tooltip. (not in this feature currently, let me know some of your thoughts here please)

This feature also is a proof of concept to have a site-wide refined prun context menu for various commands that could be edited along side the left sidebar buttons in the gameplay tab. (not in this feature though) Would you be interested?

This feature also puts some information into user data with a migration. The selected exchange is what I wanted to save. I've tested it between versions as I could, and it works, but maybe there is more to account for than what I could see at the time. Let me know if the save file migration for adding this option works as intended please.

Some pictures for reference:

<details>

Capture4
Capture
Capture2
Capture3

</details>


Razenpok:

Razenpok:
Oh man, do I procrastinate reviewing this PR. My thoughts so far:

- Adding a context menu for material icons is a great idea
- The general flow for a paying PRO player is just opening a `MAT` buffer and then clicking `CXM` in the context bar to access all CX functions. Adding a context menu might improve this flow, but I wouldn't expect players to readily switch from CXM muscle memory. Will be a great addition for FREE players though.
image
- It should be implemented either as a tooltip *or* as a context menu. I don't think it's a good idea to add more settings unless truly necessary
- "Site-wide refined prun context menu for various commands" sounds good, but we need valid use-cases for other places in the first place
- I don't quite like the visual implementation of this context menu and I can't quite put my finger on what doesn't look right. I'll come back to this PR once I will have a better idea for how it should look like.

PR #65, opened 2025-03-09
Feature prod-collapse-orders

CoroNaut:
~Draft for right now.~

Having to scroll down to see the queue of a production line is always a hassle, and I need to look there often to make sure I don't let my queue go empty. This feature asks you how many active and queued orders you want to view and then collapses the rest. After fiddling a little bit with how many orders you want to hide, you can take the scroll bar out of the equation forever! Whether you only want to see the queue, or only want to see the next 5 orders to complete, this feature and PR adds that ability.

This feature acts only on the `PROD` buffer and saves the amount of orders you want hidden to the tile information using `computedTileState` like `minimize-headers`.

A few pictures:
<details>

Capture
Capture2
Capture3
Capture5

</details>


Razenpok:
Hi, I don't have time today to properly look into this PR, but I can tell for sure that you should not use `onApiMessage`. There must be a way to properly implement what you want with `computed` and `watchEffect` (or whatever else from the Vue reactive toolkit)

CoroNaut:
I thought about using a mutation observer to just watch the column but figured using `onApiMessage` was much better for performance. I'll take a look into into watching for new buildings some other way. Is there a specific reason `onApiMessage` shouldn't be used for features? Is there some overhead there or something?

Razenpok:
`onApiMessage` is a low-level API used in the `infrastructure/prun-api` to manipulate entity stores. Since all data from API messages ends up in those entity stores, there is 100% a way to derive data (using `computed`) or behavior (using `watchEffect`) you need from them.

Razenpok:

Razenpok:
> I just want to test whether this reacts correctly when a new building is constructed or a new production line is made and then ill take it out of draft.

Let me know when it's ready for a review

CoroNaut:
New buildings are properly handled and I did some code style stuff too. Took it out of draft.

src/features/advanced/prod-collapse-orders/prod-collapse-orders.ts
```diff
@@ -0,0 +1,128 @@
...
+      const result = { ...hideOrdersInfo.value };
+      if (data[1] === -1 && data[2] === -1) {
+        delete result[data[0]];
+      } else {
+        result[data[0]] = [data[1], data[2]];
+      }
+      hideOrdersInfo.value = result;
+    },
+  });
+
```
erendrake: The getter returns the full `hideOrdersInfo` object, but the setter expects a `[string, number, number]` tuple. This asymmetry makes the API confusing to follow.

Consider replacing the setter with plain helper functions like `setHideInfo(type, keepProd, keepQueue)` and `clearHideInfo(type)` that mutate `hideOrdersInfo` directly. That would make the intent clearer at each call site.

PR #70, opened 2025-03-12
`timestampEachSecond` removal in favor of using minutes or hours

### LINKED ISSUE #67: Trim down usage of `timestampEachSecond`
### Description

The computed variable `eta` inside `prod-order-eta` using `timestampEachSecond.value` is calculated every second. It should instead do the calculation for a new eta only when a new order is placed/removed or the production line capacity changes, etc.

https://github.com/refined-prun/refined-prun/blob/6232e18307b28d8ac540fd3aad9438aa6efc0894/src/features/basic/prod-order-eta.ts#L31-L37

Simply put a console.log("eta") inside the eta computed variable to see it spam the console.

`eta` is computed every second, ~and because `completion` is computed, `completion` is also computed every second~. This means that for every `C.OrderSlot.container` in any open `PROD` buffer, they are all computing their eta time every second. `eta` should be a one-and-done calculation instead.

I'm unsure if other files containing `timestampEachSecond` are computed each second, but it should probably be checked out. ~This is an big performance bug.~

<details>

Image

</details>

Simply have a `PROD 34343434` buffer open.

### Extension version

25.3.8

### Browser(s) used

firefox

CoroNaut:
Close #67 

Many tiles in prun are calling `timestampEachSecond` every second, which is a bit much, this can be cut down at least a little. I've changed the update times of each implementation as such:
1. Balance sheets from seconds to minutes
2. Contract condition from seconds to minutes (I considered hours, but this might trip up people near a deadline)
3. contract conditions generally from seconds to minutes (works with balance sheet stuff, not super critical)
4. `XIT HEALTH` from seconds to minutes (I've checked a few times and the CX updated generally every 10 minutes anyways)
5. `XIT REP` from seconds to hours (buildings decay pretty slowly lol, and the calculations are done when opening the window too)
6. eta calculations from seconds to Date.now() (The idea of an eta is to get a constant, in-the-future time for the completion of a task, which can be calculated immediately. Only checking dependencies for updates should keep the eta accurate.)

In response to your comment https://github.com/refined-prun/refined-prun/issues/67#issuecomment-2715808839
I checked out `prod-order-eta` with this:
```return `(${formatEta(timestampEachSecond.value, completion.value)})\n(${formatEta(Date.now(), completion.value)})`;```
It didn't change any of the eta calculations as far as I could tell, picture below. When queued orders move up, they all re-calculate as well by themselves which is fine.

I took out `timestampEachSecond` from dayjs because nothing required critical second updates as far as I could tell. I've added `timestampEachHour` and used it a little. Let me know if this is decent, I can change stuff around if you want.

Also, my vscode got a bit too happy when organizing imports, they should be all 100% the same except for the timestamp imports.

<details>

Capture

</details>

### DIFF
--- a/src/core/balance/balance-sheet-live.ts
+++ b/src/core/balance/balance-sheet-live.ts
@@ -1,30 +1,30 @@
+import { timestampEachSecond } from '@src/utils/dayjs';
+import { currentAssets } from '@src/core/balance/current-assets';
+import { nonCurrentAssets } from '@src/core/balance/non-current-assets';
+import { currentLiabilities } from '@src/core/balance/current-liabilities';
+import { nonCurrentLiabilities } from '@src/core/balance/non-current-liabilities';
 import { PartialBalanceSheet } from '@src/core/balance/balance-sheet';
 import {
   calcAcidTestRatio,
+  calcLiquidationValue,
   calcDebtRatio,
   calcEquity,
-  calcLiquidationValue,
   calcQuickAssets,
   calcQuickLiabilities,
   calcTotalAssets,
   calcTotalCurrentAssets,
   calcTotalCurrentLiabilities,
-  calcTotalIntangibleAssets,
   calcTotalLiabilities,
   calcTotalNonCurrentAssets,
   calcTotalNonCurrentLiabilities,
+  calcTotalIntangibleAssets,
 } from '@src/core/balance/balance-sheet-summary';
-import { currentAssets } from '@src/core/balance/current-assets';
-import { currentLiabilities } from '@src/core/balance/current-liabilities';
-import { nonCurrentAssets } from '@src/core/balance/non-current-assets';
-import { nonCurrentLiabilities } from '@src/core/balance/non-current-liabilities';
-import { timestampEachMinute } from '@src/utils/dayjs';
 
 export const liveBalanceSheet = createLiveBalanceSheet();
 
 function createLiveBalanceSheet(): PartialBalanceSheet {
   return unwrapRefProperties({
-    timestamp: timestampEachMinute,
+    timestamp: timestampEachSecond,
 
     assets: {
       current: unwrapRefProperties({
--- a/src/core/balance/balance-sheet-live.ts
+++ b/src/core/balance/balance-sheet-live.ts
@@ -1,30 +1,30 @@
-import { timestampEachSecond } from '@src/utils/dayjs';
-import { currentAssets } from '@src/core/balance/current-assets';
-import { nonCurrentAssets } from '@src/core/balance/non-current-assets';
-import { currentLiabilities } from '@src/core/balance/current-liabilities';
-import { nonCurrentLiabilities } from '@src/core/balance/non-current-liabilities';
 import { PartialBalanceSheet } from '@src/core/balance/balance-sheet';
 import {
   calcAcidTestRatio,
-  calcLiquidationValue,
   calcDebtRatio,
   calcEquity,
+  calcLiquidationValue,
   calcQuickAssets,
   calcQuickLiabilities,
   calcTotalAssets,
   calcTotalCurrentAssets,
   calcTotalCurrentLiabilities,
+  calcTotalIntangibleAssets,
   calcTotalLiabilities,
   calcTotalNonCurrentAssets,
   calcTotalNonCurrentLiabilities,
-  calcTotalIntangibleAssets,
 } from '@src/core/balance/balance-sheet-summary';
+import { currentAssets } from '@src/core/balance/current-assets';
+import { currentLiabilities } from '@src/core/balance/current-liabilities';
+import { nonCurrentAssets } from '@src/core/balance/non-current-assets';
+import { nonCurrentLiabilities } from '@src/core/balance/non-current-liabilities';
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/features/basic/prod-order-eta.ts [hunk 1]
```diff
@@ -33,7 +32,7 @@ function onOrderSlotReady(slot: HTMLElement, siteId: string) {
       return undefined;
     }
 
-    return `(${formatEta(timestampEachSecond.value, completion.value)})`;
+    return `(${formatEta(Date.now(), completion.value)})`;
```
Razenpok, line 35: As I said in #67, you need an up-to-date timestamp to decide if this `+Xd` is needed in eta. For example:
image
Here, when your local time gets to 6:56 PM (23h 59m before the target date), eta based on `timestampEachSecond` will change to `6:55 PM`, while eta based on `Date.now()` will remain at `6:55 PM +1d`.

You can change it to `timestampEachMinute`, but the timestamp should stay reactive to correctly display the eta.

src/features/basic/sfc-flight-eta.ts
```diff
@@ -28,7 +27,7 @@ function onRowReady(
     getFlightSegmentArrival(ship.value, firstColumn.value, planId.value),
   );
   const eta = computed(() =>
-    arrival.value ? ` (${formatEta(timestampEachSecond.value, arrival.value)})` : undefined,
+    arrival.value ? ` (${formatEta(Date.now(), arrival.value)})` : undefined,
```
Razenpok, line 30: Same problem as in prod-order-eta

src/features/basic/flt-arrival-eta.ts
```diff
@@ -18,7 +17,7 @@ function onRowReady(row: HTMLTableRowElement) {
     return flight?.arrival.timestamp;
   });
   const eta = computed(() =>
-    arrival.value ? ` (${formatEta(timestampEachSecond.value, arrival.value)})` : undefined,
+    arrival.value ? ` (${formatEta(Date.now(), arrival.value)})` : undefined,
```
Razenpok, line 20: Same problem as in prod-order-eta

src/features/XIT/HEALTH.vue
```diff
@@ -36,7 +36,7 @@ const otherData = computed(() => [
   ['CXOS', cxosStore.all.value?.length],
   ['FXOS', fxosStore.all.value?.length],
   ['Currency', (balancesStore.all.value?.length ?? 0) > 0],
-  ['Last CX Price Update', cxStore.fetched ? `${dayjsEachSecond.value.to(cxStore.age)}` : false],
+  ['Last CX Price Update', cxStore.fetched ? `${dayjsEachMinute.value.to(cxStore.age)}` : false],
```
Razenpok, line 39: With this you will probably see `in a few seconds` in the `Last CX Price Update` field for a minute before the price update happens. Try reducing the cx price update interval to 3 minutes and see how it goes with `dayjsEachMinute`

src/features/XIT/CONTC/ConditionRow.vue
```diff
@@ -14,10 +14,11 @@ const eta = computed(() => {
   if (!isFinite(deadline)) {
     return '∞';
   }
-  if (deadline <= timestampEachSecond.value) {
+  const timestamp = timestampEachHour.value;
```
Razenpok, line 17: This will straight up not work when the countdown goes down to seconds
image
The requirement is to show "Xm 59s" => "Xm 58s" => "Xm 57s" => "Xm 56s" => etc. If you use `timestampEachHour`, the countdown will freeze for an hour

src/core/balance/balance-sheet-live.ts
```diff
@@ -1,30 +1,30 @@
...
+import { nonCurrentAssets } from '@src/core/balance/non-current-assets';
+import { nonCurrentLiabilities } from '@src/core/balance/non-current-liabilities';
+import { timestampEachMinute } from '@src/utils/dayjs';
 
 export const liveBalanceSheet = createLiveBalanceSheet();
 
 function createLiveBalanceSheet(): PartialBalanceSheet {
   return unwrapRefProperties({
-    timestamp: timestampEachSecond,
+    timestamp: timestampEachMinute,
```
Razenpok, line 27: The `timestamp` of the `liveBalanceSheet ` is not used anywhere in the UI, why do you want to change it to `timestampEachMinute`?

Razenpok (changes requested):
I think you should disable imports auto-sorting in your editor. Otherwise, we will start getting merge conflicts if one file is touched in two branches at the same time. Sorting imports is good, but it should be implemented project-wide (with eslint or prettier)

src/features/basic/prod-order-eta.ts
CoroNaut: Ahhh I understand now, my bad there. I was dead set on the fact that the actual 6:55pm never changes, it didn't click that the +1d could possibly roll over. +1d is too far in the future for me lol.

CoroNaut:
I was thinking the timestampEachSecond in the balance sheet live was another computed second-interval update thing you put in to keep it up to date like the order eta's. The computed stuff is turning out to be my bane, I'll eventually get it.

Razenpok (approved):

PR #72, opened 2025-03-12
Feature `prodco-order-eta`

CoroNaut:
Seeing what time an order will complete before you post it would very handy. This PR makes it possible. It grabs the template from the order window, grabs the order size value from the slider, and calculates the time it will complete and puts that just below the duration. Its reactive to changes in the template, order size, and anything the current production line might do. 

I couldn't find any cool template ID or anything similar in HTML that would streamline getting the template duration, so I made a custom function for getting the template which turned out pretty good. 

A few pictures of putting the order into the queue:
<details>

Capture
Capture2

</details>


Razenpok:
@CoroNaut fyi, I've pushed some changes related to readability. Please note how `watch` can be replaced with `computed` values.

CoroNaut:
Oh, I didn't realize `const line = computed(() => productionStore.getById(tile.parameter)!);` would work as a computed. I thought that it was simply calling a function and returning a regular non-ref value. I used `productionStore.all` because it was immediately a ref. I'll try to use that in the future.

Razenpok:
Yep, under the hood `getById` touches `.all`, which is also reactive. All prun api stores have the ` getById` function, and most have additional functions, like `contractsStore.getByLocalId`

PR #73, opened 2025-03-13
Feature progress-bar-tooltips

CoroNaut:
Someone in global chat had the idea that it would be nice to have tooltips on the progress bars. A tooltip helps compare planets resources in `PLI` or `SYSI`. This PR simply adds the title attribute where necessary with reactive components to whatever the value is behind the scene. I just went ahead and made tooltips for the following:

1. Progress bars (like `SYSI`)
2. Factor bars (like `PLI` soil fertility)
3. Sliders (like `SFC` or `PRODCO`)
4. Orders (like `PROD`)

The environment bars in `PLI` are excluded, they aren't really a progress bar and mainly are just there for show. You really only care about them if they are in the green or in the red.

I considered putting a tiny xx% inside the progress bars but it needs like 8px font (maybe 9px) which is getting too small and its squished in there. Also coloring the text to make it contrast what color its on top of would be a hassle and probably not even look good.

### DIFF
--- a/src/features/basic/index.ts
+++ b/src/features/basic/index.ts
@@ -45,6 +45,7 @@ import './prevent-delete-button-misclicks';
 import './prod-burn-link';
 import './prod-order-eta';
 import './prodq-queue-load';
+import './progress-bar-tooltips';
 import './prun-bugs';
 import './rprun-version-label';
 import './screen-tab-bar/screen-tab-bar';
--- a/src/features/basic/progress-bar-tooltips.module.css
+++ b/src/features/basic/progress-bar-tooltips.module.css
@@ -1,3 +1,0 @@
-.progressBarPrimary {
-  pointer-events: none;
-}
--- a/src/features/basic/progress-bar-tooltips.module.css
+++ b/src/features/basic/progress-bar-tooltips.module.css
@@ -1,0 +1,3 @@
+.progressBarPrimary {
+  pointer-events: none;
+}
--- a/src/features/basic/progress-bar-tooltips.ts
+++ b/src/features/basic/progress-bar-tooltips.ts
@@ -1,0 +1,68 @@
+import { applyClassCssRule } from '@src/infrastructure/prun-ui/refined-prun-css';
+import { percent2 } from '@src/utils/format';
+import { refAttributeValue } from '@src/utils/reactive-dom';
+import classes from './progress-bar-tooltips.module.css';
+import { watchEffectWhileNodeAlive } from '@src/utils/watch';
+
+function init() {
+  applyClassCssRule(C.ProgressBar.primary, classes.progressBarPrimary);
+
+  subscribe($$(document, C.ProgressBar.container), progressBar => {
+    // FLT has primary bar and secondary bar.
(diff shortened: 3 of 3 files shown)


Razenpok:
You don't need to repeat code from `watchEffectWhileNodeAlive` - under the hood it uses Vue `watchEffect` which executes the parameter function immediately

https://vuejs.org/guide/essentials/watchers.html#watcheffect

I also recommend reading the whole article for bonus knowledge points

CoroNaut:
Yep there it is `Here, the callback will run immediately... `. I was looking at `watchEffectWhileNodeAlive` and thought I had to set `WatchOptionsBase` to get it to run immediately, then I found out only `WatchOptions` has the `immediate` flag to run the code immediately, so I gave up there.

Razenpok:
> Someone in global chat had the idea that it would be nice to have tooltips on the progress bars.

I think they might've meant the actual tooltip, like in ⓘ icons. iirc the tooltip of elements with "title" shows up after some delay, so this feature will probably be unseen by most players. Do you see yourself using this feature?

CoroNaut:
I'm definitely certain they were talking about tooltips on the progress bars to show their value. I remember they were talking specifically about PLI and SYSI resource bars. The interface just shows a bar, so comparing two bars visually for which one is higher or lower is not even worth doing. It basically requires someone to go to third party tools to compare resources unless you just wing it every time for which planet is better. A 2-3 minute journey just to lookup resource values for 2 planets is kind of annoying when they could just have the value in-game, other than a bar.

For me specifically, I use third party tools for mass comparisons of planets, so it won't be often that I use it. If I had just a quick thought to compare two planets for some reason, having the tooltips is definitely handy. Just open up the PLI real quick and hover, "oh, 35.55% ... and .. 37.74% for water. Yup, that one is better". Otherwise I probably wouldn't care enough to go on the 2-3 minute journey to compare 1 resource on two planets using third party tools.

If the main worry is that it will go unseen, I could just make an additional text beside the bars only in PLI and SYSI:

Capture

Razenpok:
I'm not sure if we are on the same page, so here's a video:

https://github.com/user-attachments/assets/90708fa5-e462-4efd-be04-f90bb32996a9

In the video, you can see that both tootips show the value of the progress bar.
The first tooltip shows instantly and looks like the one in ⓘ icons. This tooltip is added via the `data-tooltip` attribute.
The second tooltip is a system tooltip that takes ~2 seconds to appear. This tooltip is added via the `title` attribute.

The problem with the second tooltip is that (in my opinion) most people will never see it, since the common pattern in PrUn is that tooltips show up instantly.

CoroNaut:
Yeah that's my bad. Sometimes I'm just that blind. I completely forgot data-tooltip existed. I like the look of the data-tooltip and converted everything to use them. I changed up the css of the tooltips to be a bit more slim instead of being padding:8px, which makes a huge tooltip box. The little arrow on the border of the data-tooltip wasn't playing nice and sometimes it was 3px away or 1px inset, so I just removed it.

<details>

Capture
Capture2
Capture3
Capture4

</details>

PR #76, opened 2025-03-17
Feature `prodq-order-eta`

CoroNaut:
The `prod-order-eta` code can mostly be re-used for `PRODQ`. This PR just changes how the elements are selected, the actual time calculation is the same. I also put a css rule on the 'Completion / Duration' to shorten the length of the overall column since it was causing the entire column to stretch.

If only the `PRODQ` window was sorted like `PROD`.

`calcCompletionDate` is duplicated from `prod-order-eta`, I can move it if you have a location in mind, then export it if you're interested.

Before and after pictures:
<details>

Capture
Capture2

</details>


src/features/basic/prodq-order-eta copy.module.css
```diff
@@ -0,0 +1,3 @@
+.thCompletion {
+  width:min-content;
```
Razenpok, line 2:
```suggestion
  width: min-content;
```

src/features/basic/prodq-order-eta.module.css
```diff
@@ -0,0 +1,3 @@
+.thCompletion {
+  width: min-content;
```
Razenpok, line 2: What's the purpose of this change? I see no difference after removing it

src/features/basic/prodq-order-eta.ts
```diff
@@ -0,0 +1,57 @@
...
+        return;
+      }
+      onOrderSlotReady(order.children[5] as HTMLElement, orderId, tile.parameter!);
+    });
+  });
+}
+
+function onOrderSlotReady(
+  slot: HTMLElement,
+  orderId: globalThis.Ref<string | null, string | null>,
```
Razenpok, line 25:
```suggestion
  orderId: Ref<string | null>,
```

Razenpok (changes requested):

src/features/basic/prodq-order-eta.module.css
CoroNaut: With it, the header row will always be two lines, which makes the row about 10px more in height, but allows the column to instead scale to the rest of the rows in that column. This means that the entire table will flex more naturally so all columns get to scale to the contents of the row orders, which makes things much more uniform and looks better.

Without it, the 'completion / duration' text always tries to widen the column as much as possible, even when its text can't fit on one line. This artificially increases the width of the entire column solely based on trying to fit the 'completion / duration' text. This looks kind of ugly IMO.

Capture
Capture2

Razenpok: Huh, I guess it's a Firefox-only behavior since I don't observe it in Chrome

PR #77, opened 2025-03-17
Feature `header-hide-controls-button`

CoroNaut:
Sometimes the context controls just add additional space when you don't use them anyway. This PR makes a button in the header of every buffer with context controls that is a little eye. Click that and your context controls will vanish or reappear. The state is saved in user data just like `minimize-headers`.

I put this feature beside the other two header buttons in `BASIC` features.

Before and after pictures:
<details>

Capture
Capture2

</details>


Razenpok:
How about this:
- When the context bar is visible, it has a crossed-eye button inside the context bar. You can create a gap by creating a `flex-grow: 1` element between buttons)

image

- When the context bar is hidden, you show a bookmark button or some other icon that fits the meaning of the button

https://fontawesome.com/icons/bookmark?f=classic&s=solid
image


This way, the feature won't clutter tile controls when the context bar is not hidden, which will likely be the case for most tiles.

src/features/basic/header-hide-controls-button/header-hide-controls-button.ts
```diff
@@ -0,0 +1,38 @@
+import { createFragmentApp } from '@src/utils/vue-fragment-app';
+import TileControlsButton from '@src/components/TileControlsButton.vue';
+import { computedTileState } from '@src/store/user-data-tiles';
+import { getTileState } from './tile-state';
+import { watchEffectWhileNodeAlive } from '@src/utils/watch';
+
+async function onTileReady(tile: PrunTile) {
+  const tileContextControls = await $(tile.frame, C.ContextControls.container);
+  if (!tileContextControls) {
```
Razenpok, line 9: This check is unnecessary since `await $` will always produce non-nullish values

src/features/basic/header-hide-controls-button/header-hide-controls-button.ts
```diff
@@ -0,0 +1,38 @@
...
+    return;
+  }
+
+  const isMinimized = computedTileState(getTileState(tile), 'minimizeContextControls', false);
+  watchEffectWhileNodeAlive(tile.anchor, () => {
+    if (isMinimized.value) {
+      tileContextControls.style.display = 'none';
+    } else {
+      tileContextControls.style.display = 'flex';
+    }
```
Razenpok, line 19: Replace with addition and removal of the `css.hidden` class from utils

src/features/basic/header-hide-controls-button/tile-state.ts
Razenpok, line 1: Merge this file into the feature file. The feature file is small enough, so it's not a problem to have this code there.

Razenpok (changes requested):

Razenpok:
image

Context controls look funny now

CoroNaut:
I changed up the context controls file as a first attempt at solving the problem, but then decided against it and forgot to revert the file. I do have it making a separate component in the PR and I reverted that file.

Razenpok:
Made some more changes:
1. Removed `css.HIDDEN` in favor of `isMinimized.value ? (TileControlsButton) : null`. You can conditionally render elements like this if writing JSX/TSX.
2. Replaced unicode characters with unicode values. In the future, always prefer unicode values, as they are easier to search for.
3. Inlined the `flex-grow: 1` div into the `createFragmentApp` call. Notice the `<>` syntax - that's a template. You can read about it here: https://vuejs.org/api/built-in-special-elements.html#template. In .tsx files they are written as `<></>`

PR #78, opened 2025-03-20
Feature `nots-reactive-format`

### LINKED ISSUE #9: Move Notification Label to free space
### Description

Currently, on small buffers the notification type takes a lot of space. An idea would be to move it above the notification text / let it float up top left:

Screenshot 2024-11-28 at 15 25 55

CoroNaut:
Closes #9

The `nots-notification-type-label` makes some dead space on the left side of notifications which can make a large blank space if the `NOTS` window is small. This PR adds a new feature that moves the date to the top next to the type label. This puts the dead space at the top right instead. In addition, it has a dependency that requires `notes-notification-type-label`. When `notes-notification-type-label` is disabled, this feature, `nots-change-format` will subsequently be disabled before it is loaded as well. 

In the after picture, buffer 8 is just a little smaller than buffer 7 to show what it looks like

Before and after pictures:
<details>

Capture2
Capture4

</details>


Razenpok:
Merge this feature into `nots-notification-type-label`, as vanilla notifications don't need dynamic layout

image

Razenpok:
And for the future: if you need another feature as a dependency, there's 95% you are doing something wrong

src/features/basic/nots-reactive-format.module.css
```diff
@@ -0,0 +1,27 @@
...
+.alertListItemInnerContent {
+  margin-left: 0px !important;
+}
+
+.alertListItemInnerContentBefore {
+  content: '';
+  display: block;
+  float: left;
+  height: 15px;
+  width: calc(79.99px + max(calc(calc(100% - 280px) * -1000000),0px));
```
Razenpok, line 16: Please explain what the math is going on here

src/features/basic/nots-reactive-format.module.css
```diff
@@ -0,0 +1,27 @@
+.alertListItemContent {
+  position: relative;
+  overflow-x: hidden;
+  min-height: 34.2px;
+}
+
+.alertListItemInnerContent {
+  margin-left: 0px !important;
```
Razenpok, line 8: Why do you need !important here?

src/features/basic/nots-reactive-format.module.css
CoroNaut: There is a ::before invisible box that keeps the notification inner text from overlapping the time/label. Notice the purple line, the X-axis is the width of the individual notification element in this case. The `calc` makes it so the width of the invisible div goes from 80px to >=300px if the width of the inner text box (its parent) goes from >=300px to <300px. This invisible div stops the inner text from overlapping the label/time because both of those have absolute position. When the window is wide, it stops the inner text from going left, when slim, it stops the inner text from going up.

The width of the `NOTS` window can only be in whole-pixel numbers, so 300px -> 299px and no in-between. The elements shift between this transition.

Capture

The `left` in `alertListItemTime` is similarly calculated to make the `left` value is 10px when the element is >299px wide, and 80px when <=299px wide. This moves the time from directly below the notification label to the right of the label.

Capture2

The `top` in `alertListItemTime` is calculated to make sure that if the height of the inner text box exceeds 35px, the `top` goes from 18px to 6px. This moves the time from below the notification label, to the same line as the label.

The website is desmos.com.

src/features/basic/nots-reactive-format.module.css
CoroNaut: It was to counteract 
https://github.com/refined-prun/refined-prun/blob/1cc4acf4288bc17e2099027582c974b192f0d026/src/features/basic/nots-notification-type-label.tsx#L29-L30

Razenpok:
Whew, the math is scary. What do you think about this simpler approach?

1. Add `display: flex; flex-wrap: wrap;` to `AlertListItem__content___XYebR1E`
2. Remove `position: absolute;` from `rp-nots-notification-type-label__label___dac599b`
3. Remove `margin-left: 65px;` from the `span` with the notification text. You can also remove `display: block;` I guess? It doesn't seem to do anything inside `display: flex` `AlertListItem__content___XYebR1E`.
4. Add `width: 100%;` to `AlertListItem__time___xNtrNGc`

Take a look at the first `ARRIVAL` notification (surgically added the changes in DevTools):

https://github.com/user-attachments/assets/fe9c4473-20f1-4651-a6d1-8f303e09b1b8

CoroNaut:
I can't really see any way to have the sentence stay to the right and flexibly take up two lines and still keep the label and time vertically adjacent. Thats fine though, your approach works just fine for the use case.

Capture

Razenpok:
I've changed the `flex` of the text span to `flex: 1 1 200px;`, so now it will shrink and wrap text until it's 200px wide, and then span itself will wrap under the type label.

image

Razenpok:
Side note: I'm grateful that you include detailed descriptions and screenshots in your PRs. Thanks!

PR #79, opened 2025-03-22
Feature `prodq-hide-government-links`

CoroNaut:
Extracted from #75. This PR adds a feature that hides the 'Government of Hortus - Promitor' link in the fee column.

Picture with the feature on:
<details>

Capture

</details>


src/features/advanced/prodq-hide-government-links.ts
```diff
@@ -0,0 +1,8 @@
+import { applyScopedClassCssRule } from '@src/infrastructure/prun-ui/refined-prun-css';
+import css from '@src/utils/css-utils.module.css';
+
+function init() {
+  applyScopedClassCssRule('PRODQ', C.Link.link, css.hidden);
```
Razenpok, line 5: Right, so this will definitely work, but let's err on the safe side and scope it more specifically. Something like:
```css
tr td:nth-child(3) .${C.Link.link}
```

src/features/advanced/prodq-hide-government-links.ts
```diff
@@ -0,0 +1,8 @@
...
+}
+
+features.add(import.meta.url, init, 'PRODQ: Hides government links.');
```
Razenpok, line 8:
```suggestion
features.add(import.meta.url, init, 'PRODQ: Hides fee collector links.');
```
More specific

Razenpok (changes requested):

PR #80, opened 2025-03-23
Feature `cx-search-bar`

CoroNaut:
I'm a bit tired of having to scroll through the cx to find what I want to buy or sell. The text is a bit small and a little tough on the eyes. To top it off, they are sorted by their back-end material name, not even sorted by ticker or their pretty name that is actually displayed. This PR makes a search bar and 'Results only' button in the CX window. 

Usage: Enter any text into the search bar and immediately the list will tint your matching materials green, as well as categories in the drop-down that have matches. Turn on 'Results only' and all rows that don't matter will vanish, leaving only relevant results. This also works for the drop-down list for categories. Enter your text and choose from a smaller list of categories that contain results, and view the materials. No more scrolling through categories or materials!

I generated an object with all the current tickers and their pretty names since I couldn't find them in any material objects. I'm not sure if you want this kind of data file somewhere specific, or if the `cx-search-bar` folder is fine.

Some pictures:
<details>

Capture
Capture2
Capture3
Capture4
Capture

</details>


Razenpok:
1. What's the use case for a disabled "Results only" toggle?
2. Localized material names are available via a `getMaterialName` function from `i18n.ts`. You can see how it's used in the `nots-material-ticker` feature.

CoroNaut:
1. I haven't really thought about it. I like user-choice settings and it seemed natural to include. A user just looking for a trade button or quick market prices would probably always want it toggled on. With the toggle off, some users can still just browse the cx with some results pointed out and look at nearby items that aren't in the search. They would have to empty the search bar every time they wanted to browse again. Without the toggle, I suppose the green coloring can go. I'm not really leaning either way on it. Lemme know if you think it would go unused or be a pain to turn on every time. Alternatively, it could default to 'on'.

2. Neat!

src/features/advanced/cx-search-bar/cx-search-bar.module.css
```diff
@@ -0,0 +1,25 @@
...
+
+  tbody tr:nth-child(2n+1)::after {
+    content: "";
+    position: absolute;
+    inset: 0px;
+    pointer-events: none;
+  }
+}
+
+.textInputElement {
```
Razenpok, line 22: Uhh, we got a misunderstanding here. If you use a class in a `.vue` file, it should be inside the `<style module>` section of such `.vue` file

CoroNaut: I had textInputElement inside the css module and definitely intended to keep it there, but when I built it, the $style inside the vue template wasn't applying. The import $style inside the script section was getting priority over the `$style.textInputElement` from the vue style module.

I'll put those styles inside the vue component and use `const $style = useCssModule();` instead, that makes much more sense lol.

CoroNaut: I separated the css into their correct files and built it and it works as intended, ty.

Razenpok:
TL;DR: Let's remove the "Results only" toggle and green highlight

While I get the appeal of giving users choices, I have a different approach to this extension: I try to add as few new elements as possible because PrUn UI is already packed with information. There are several benefits to it:
1. Less UI bloat
2. Less code to maintain
3. Adding new controls after receiving player feedback is easier than removing extra controls that found almost no use in the player base, as there's a chance to piss someone off who _will_ end up using them

You might also notice that `XIT SET` has almost no feature settings - that's by design. Features should either:
- Work for everyone as a nonconfigurable default, or  
- Have settings placed right where the feature is used

Another thing is that all features in the extension are enabled by default. "Disabled by default" means most players don't need it, so it doesn't belong to rprun.

So, yeah, it's about adding things that people will _definitely_ use, and here I don't see how most players would use the "Results only" toggle. That said, if we see real demand for it later, we can always add it back.

CoroNaut:
Removes results only and added just a button with X on it to clear search results so the user doesn't have to click into the search bar and ctrl-a and delete the text.

src/features/advanced/cx-search-bar/cx-search-bar.ts
```diff
@@ -0,0 +1,22 @@
+import MaterialSearchAndResults from './MaterialSearchAndResults.vue';
+import $style from './cx-search-bar.module.css';
+
+function onTileReady(tile: PrunTile) {
+  subscribe($$(tile.anchor, C.ComExPanel.input), comExPanel => {
+    const actionBar = _$(comExPanel, C.ActionBar.container)!;
```
Razenpok, line 6:
```suggestion
  subscribe($$(tile.anchor, C.ComExPanel.input), async comExPanel => {
    const actionBar = await $(comExPanel, C.ActionBar.container);
```

Use `async/await $` where possible, as this function is type-safe and requires no `!` operator.

Razenpok (changes requested):
Extract all external element handling from the Vue component into the feature file. This way, this feature will be consistent with how all other features are written, and your Vue component won't be overloaded with irrelevant logic. Use callbacks in props to communicate values/events from Vue to the feature.

CoroNaut:
Since the vue template was simple enough, I just put it into a tsx file.

src/features/basic/cx-search-bar.tsx
```diff
@@ -0,0 +1,92 @@
...
+  watchEffectWhileNodeAlive(comExPanel, () => {
+    const searchTerm = searchText.value.toUpperCase();
+
+    categoryOptions.forEach(resetMatches);
+    materialRows.forEach(resetMatches);
+
+    const materials = materialsStore.all.value;
+    if (searchTerm.length === 0 || !materials) {
+      return;
+    }
```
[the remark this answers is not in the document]
Razenpok: Create an issue for this suggestion. Add a note that these improvements can be also applied in inv-search

PR #81, opened 2025-03-24
Add "hide timestamps" to `hide-system-chat-message`

CoroNaut:
I have small chat windows on my window because I like to look every so often and keep up with the buzz. The timestamps take up a lot of space for no reason. If you really need to know the time a message was sent, just click the button to view them. This PR adds it to `hide-system-chat-message` as a toggle. I considered making it a separate feature, but the user options were right there already.

I also changed the .remove and .add functions into a single toggle:
```
      messages.classList.remove(classes.hideJoined, classes.hideDeleted);
      if (hideJoined.value) {
        messages.classList.add(classes.hideJoined);
      }
```

<details>

Capture
Capture2
Capture3

</details


Razenpok:
What do you think about this: nice or meh?

image

CoroNaut:
Well, I do like that its more compact. It may look a little weird, but the eyes will mainly be at the bottom of the window at any given point anyway.

The only thing I can think of to help would be a button to hide the whole bar like the context controls hide feature. The options to show/hide, User list, mute... all are pretty much one-and-done buttons for the most part, a feature to hid them wouldn't be a stretch. I may have clicked 'User list' once and mute never. "Hide the hide controls" has a funny sound to it.

Capture

Razenpok:
I've asked around in Discord, and this variant is deemed best:

image
Notice the inversion of toggles: "hide X" => "X"

Hiding the filter bar sounds like a good idea. Let's use a different icon to avoid confusing it with the `header-hide-controls-button`, like https://fontawesome.com/icons/filter?f=classic&s=solid
To hide the bar, add a fa xmark to the right side of the filter bar: https://fontawesome.com/icons/xmark?f=classic&s=solid

Before merging, we'll rename this feature to something different, like `chat-filters` (accepting better ideas for the name)

CoroNaut:
I inverted toggles, added hide/show buttons, and renamed file.

Capture

Razenpok:
Oof, you've been too quick with renaming the feature - now the diff is all green, and it's difficult to say which lines are new and what was removed. Can you revert the last two commits and reapply the changes to the old files?

Razenpok:
The other thing is that you cannot just rename the files. You also need to add a user data migration to migrate the feature id in the `settings.disabled` array. That's why I said, "**Before merging**, we'll rename this feature" - to limit these unrelated changes to the very last moment.

Razenpok:
> My 'x' icon looks a little skewed on the PU website, but on font-awesome site, it isn't skewed. Might just be a firefox thing, idk.

That's because you've used the `<i>` tag in `<i class={[fa.solid]}>{'\uf00d'}</i>`. `<i>` is the HTML tag for italic, like `<b>` is for bold. To fix it, you need to add `font-style: normal;` to the `fa.solid` and `fa.regular` CSS rules.

CoroNaut:
I'm probably not understanding you right. The commits show a diff with specific line changes even in renamed files and not a 100% green diff. The 'files changed' tab will always show new files as 100% green as far as I know. I could revert the last 2 commits and make a commit that shows a diff with just text changes, but it won't build because it needs to be `.tsx` file for the 'x' and filter buttons I added. Is this what you want me to do?

I also didn't think it needed a migration since the userdata is just `disabled: [] as string[],`. Every new chat window that needs to save its data will make the data and push it into the array. For users without a migrated disabled array, the main `hide-system-chat-message` file migrates it automatically with `const hideFilterBar = computed(() => state.value?.hideFilterBar ?? false);`.

Ah, the i tag is for italics, not 'icon', gotcha. I can change that once the commit issue is resolved.

PR #85, opened 2025-03-26
Add vite temporary file to gitignore

CoroNaut:
The vite temporary file sometimes ends up in my commits which is mildly annoying. This PR adds the temporary file to gitignore so I can be forever sure I'll never have to look at it again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@
 .DS_Store
 .idea
 **/.turbo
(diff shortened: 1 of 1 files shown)


Razenpok (approved):

Razenpok:
idk why this file does not disappear after vite run, but I also occasionally see it (rarely, though)

PR #92, opened 2025-04-04
Fix `prodco-order-eta` not getting new template if specific filters are selected

CoroNaut:
The `prodco-order-eta` completion time is calculated right when opening the window or selecting any template from 'Production template'. If however the template filter is changed, there's a good chance that the prosperous universe website doesn't change enough information to trigger `observeChildListChanged`. `observeDescendantListChanged` was not enough as well. I just added `observeDescendantAttributeListChanged` that has attributes and subtree included to make sure all changes are captured. 

Pictures show that when a filter is chosen, and the template changes, the change was not enough to trigger `observeChildListChanged`
<details>

Capture
Capture2

</details>


Razenpok:
@CoroNaut the `prodco-order-eta` just stopped working completely after the maintenance update. Reason: The structure of the `ProductionLine__template___HfCtRrQ` object has been changed to the following:
image
I think this change was made to accommodate the big list of inputs for consumable bundles, which also got released today with the maintenance update.
The "DOM elements don't change on material filter change" behavior can be observed on the new `ProductionLine__inputs___bKTWKmj` element, just like on `ProductionLine__template___HfCtRrQ`, so this PR is still relevant.
If you add a fix for `ProductionLine__inputs___bKTWKmj` to this PR, I can merge it to the next release. I'm mostly done with `XIT ACT` rewrite, but I'm waiting for molp to approve them.

src/features/basic/prodco-order-eta.ts
```diff
@@ -2,15 +2,15 @@ import { productionStore } from '@src/infrastructure/prun-api/data/production';
...
+import { observeDescendantAttributeListChanged } from '@src/utils/mutation-observer';
 
 function onTileReady(tile: PrunTile) {
   const line = computed(() => productionStore.getById(tile.parameter)!);
   subscribe($$(tile.anchor, C.ProductionLine.form), form => {
     const template = ref<PrunApi.ProductionTemplate>();
     const templateField = form.children[5];
     const dropDownItem = _$(templateField, C.DropDownBox.currentItem)!;
-    observeChildListChanged(dropDownItem, () => {
+    observeDescendantAttributeListChanged(dropDownItem, () => {
```
Razenpok, line 13: Since "`C.ProductionLine.template` doesn't change on filter change" looks more like a screw up of the base game, let's just plug it with a workaround on our side. Remove the call to `observeChildListChanged` completely with this:
```typescript
    // In some edge cases C.ProductionLine.template is not re-created
    // on template change. So instead of using observeChildListChanged
    // on the dropDownItem, we'll watch its textContent, as it always
    // reflects the template change.
    const templateText = refTextContent(dropDownItem);
    watchEffect(() => {
      // Touch reactive value.
      // eslint-disable-next-line @typescript-eslint/no-unused-vars
      const _ = templateText.value;
      // The ProductionLine.template element is re-created each time
      // a new template in the drop-down box is selected.
      const templateElement = _$(dropDownItem, C.ProductionLine.template);
      template.value = parseTemplate(line.value, templateElement);
    });
```

Razenpok (changes requested):

CoroNaut:
I was able to make the template parsing a bit cleaner. I also updated `prun-css.types.d.ts` so I could use the new css class. `C.ProductionLine.inputs`

PR #123, opened 2025-06-08
Added Feature: Show Space Remaining on SHPI and INV tiles

keenan-v1:
Added remaining weight and volume to INV and SHPI tiles.

firefox_lhcpjLaW2Y

### DIFF
--- a/src/features/basic/show-space-remaining.ts
+++ b/src/features/basic/show-space-remaining.ts
@@ -1,0 +1,68 @@
+import { shipsStore } from '@src/infrastructure/prun-api/data/ships';
+import { storagesStore } from '@src/infrastructure/prun-api/data/storage';
+import { refValue } from '@src/utils/reactive-dom';
+import { watchEffectWhileNodeAlive } from '@src/utils/watch';
+
+function formatFixed(f: number) {
+  return f % 1 ? f.toFixed(2) : f.toString();
+}
+
+function getStore(tile: PrunTile) {
+  const storeId =
+    tile.command.startsWith('SHPI') && tile.parameter
+      ? shipsStore.getByRegistration(tile.parameter)?.idShipStore
+      : tile.parameter;
+  if (!storeId) {
+    return;
+  }
+  return storagesStore.getById(storeId);
+}
+
+function onTileReady(tile: PrunTile) {
+  subscribe($$(tile.anchor, C.StoreView.column), column => {
+    const capacities = _$$(column, C.StoreView.capacity);
+    if (capacities.length < 2) {
+      return;
+    }
+    const weightIndex = capacities.findIndex(x => x.textContent?.includes('Weight'));
+    const volumeIndex = capacities.findIndex(x => x.textContent?.includes('Volume'));
+
+    if (weightIndex === -1 || volumeIndex === -1) {
+      return;
+    }
+    const weightBar = capacities[weightIndex].getElementsByTagName('progress')[0];
+    const volumeBar = capacities[volumeIndex].getElementsByTagName('progress')[0];
+    if (!weightBar || !volumeBar) {
+      return;
+    }
+
+    watchEffectWhileNodeAlive(weightBar, () => {
+      const store = getStore(tile);
+      if (!store) {
+        return;
+      }
+      const weightAvailable = formatFixed(store.weightCapacity - store.weightLoad);
+      (capacities[weightIndex].lastChild as Text).textContent = ` (${weightAvailable}t)`;
+    });
+
+    watchEffectWhileNodeAlive(volumeBar, () => {
+      const store = getStore(tile);
+      if (!store) {
+        return;
+      }
+      const volumeAvailable = formatFixed(store.volumeCapacity - store.volumeLoad);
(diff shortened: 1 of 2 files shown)


src/features/basic/show-space-remaining.ts
```diff
@@ -0,0 +1,68 @@
+import { shipsStore } from '@src/infrastructure/prun-api/data/ships';
+import { storagesStore } from '@src/infrastructure/prun-api/data/storage';
+import { watchEffectWhileNodeAlive } from '@src/utils/watch';
+
+function formatFixed(f: number) {
```
Razenpok, line 5: Instead of custom format functions, use ones from the `utils/format.ts`. See an example here: https://github.com/refined-prun/refined-prun/blob/5f4872a62ddaaf31466281f4960e26175002b156/src/features/basic/cxpo-auto-price.ts#L116

src/features/basic/show-space-remaining.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+function onTileReady(tile: PrunTile) {
+  subscribe($$(tile.anchor, C.StoreView.column), column => {
+    const capacities = _$$(column, C.StoreView.capacity);
+    if (capacities.length < 2) {
+      return;
+    }
+    const weightIndex = capacities.findIndex(x => x.textContent?.includes('Weight'));
+    const volumeIndex = capacities.findIndex(x => x.textContent?.includes('Volume'));
```
Razenpok, line 28: Try to avoid using localized text; otherwise, a feature will only work in English localization. Here, either access elements by index, or use localized labels from `PrunI18N` to search for an element. I recommend the former option, for simplicity.

Razenpok: And, instead of `findIndex` use `find` - it will give you an element directly:

```suggestion
    const weight = capacities.find(x => x.textContent?.includes('Weight'));
    const volume = capacities.find(x => x.textContent?.includes('Volume'));
```

src/features/basic/show-space-remaining.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    if (weightIndex === -1 || volumeIndex === -1) {
+      return;
+    }
+    const weightBar = capacities[weightIndex].getElementsByTagName('progress')[0];
+    const volumeBar = capacities[volumeIndex].getElementsByTagName('progress')[0];
```
Razenpok, line 34:
```suggestion
    const weightBar = await $(capacities[weightIndex], 'progress');
    const volumeBar = await $(capacities[volumeIndex], 'progress');
```

src/features/basic/show-space-remaining.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (!weightBar || !volumeBar) {
+      return;
+    }
+
+    watchEffectWhileNodeAlive(weightBar, () => {
+      const store = getStore(tile);
```
Razenpok, line 40: Use a `getInvStore` function from `core/store-id.ts` instead

src/features/basic/show-space-remaining.ts
```diff
@@ -0,0 +1,68 @@
...
+      if (!store) {
+        return;
+      }
+      const weightAvailable = formatFixed(store.weightCapacity - store.weightLoad);
+      (capacities[weightIndex].lastChild as Text).textContent = ` (${weightAvailable}t)`;
+    });
```
Razenpok, line 46: To avoid issues if molp changes the node structure inside the `C.StoreView.capacity` element, append a new element instead of modifying an existing one.
```suggestion
    const weightText = computed(() => {
      const store = getStore(tile);
      return store.value ? ` (${formatFixed(store.weightCapacity - store.weightLoad)})` : undefined;
    });
    const weightSpan = createReactiveSpan(capacities[weightIndex], weightText);
    weightSpan.style.whiteSpace = 'pre';
    capacities[weightIndex].appendChild(weightSpan);
```

src/features/basic/show-space-remaining.ts
```diff
@@ -0,0 +1,68 @@
...
+      }
+      const volumeAvailable = formatFixed(store.volumeCapacity - store.volumeLoad);
+      (capacities[volumeIndex].lastChild as Text).textContent = ` (${volumeAvailable}m³)`;
+    });
+  });
+}
+
+function init() {
+  tiles.observe('INV', onTileReady);
+  tiles.observe('SHPI', onTileReady);
```
Razenpok, line 61:
```suggestion
  tiles.observe(['INV', 'SHPI'], onTileReady);
```

src/features/basic/show-space-remaining.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+features.add(
+  import.meta.url,
+  init,
+  'Show the remaining weight and volume capacity of the selected store in INV and SHPI',
```
Razenpok, line 67:
```suggestion
  'Shows the remaining weight and volume capacity of the selected store in INV and SHPI',
```

Razenpok (changes requested):

Razenpok:
I've moved the code around a bit, but otherwise, looks good to me!

Issue #122, opened 2025-06-07
Dynamic Tiles: Update tiles based on selected input


theit8514:
### Description

Back in the day this was one of my most requested features: being able to select a value and update tiles based on that selection ([example in PrUn forums](https://com.prosperousuniverse.com/t/buffer-variables/5410)).

theit8514:
I poked around at the code and found that it is possible to replace tiles with a new command using a similar method to ACT's [TileAllocator](https://github.com/refined-prun/refined-prun/blob/fd9f77ac183aab5e92b139b90a1b75f2be3d49bd/src/features/XIT/ACT/runner/tile-allocator.ts#L50). I made a small POC around changing Exchanges, Bases, or Materials. Select the desired value then click Load and the tiles update to that value. Would you accept a [PR](https://github.com/theit8514/refined-prun/tree/add-command-var) for that? Disclaimer: Vue is not my strong suit but it's functional right now.

![Image](https://github.com/user-attachments/assets/c4e7031f-1455-4b10-b799-9f45d2145ac5)

Razenpok:
Hi @theit8514, I appreciate your reference implementation.

The problem with this feature is not about code, unfortunately, but rather about Terms of Service and/or user experience. According to the Terms of Service (ToS), this entire extension can be deemed illegal and result in an in-game ban if prun developers decide to do so. Fortunately, they appreciate the value this extension brings, so the only actual line they drew is not to add features that will lead to server requests without user action, and only produce one server request per user action. This is precisely how XIT ACT works: the user needs to click the "ACT" button every time the action package wants to execute something that results in a server request, such as creating an order in CXPO, executing a transfer in MTRA, or **opening a CXPO buffer (but not MTRA)**. There are two gotchas here:
 
1. Many of the commands result in a server request upon opening. For example, CXPO will send a request for fresh order book data. However, some of the commands, such as MTRA, do not send requests upon opening.
2. Changing a docked tile always leads to a server request - that's how the game saves your screen layout. On the other hand, changing a buffer tile (one in a floating window) does not require a server request.

That's why, when executing a CX Buy action:
 
1. User clicks ACT → CXPO opens (server request)
2. User clicks ACT again → order is sent (another request)

For MTRA:
1. XIT ACT will open a floating MTRA tile for you (no server fetch)
2. User clicks ACT → transfer is executed (server request)

[...]

PR #130, opened 2025-10-07
Add XIT VAR

theit8514:
Updated per comments in #122 and the recent discord conversation to require a click for each tile change.

An observation: The UI_TILES_CHANGE_COMMAND is not getting triggered so changing the tiles does not generate any server commands, but also does not persist to the server. This may be desired to prevent many server calls per click.


theit8514:
Obsolete with the upcoming SCRN command

PR #137, opened 2025-12-13
Remember scroll

purpleorangeazure:
This feature remembers tile scroll positions and automatically applies them on tile load.

I created this because I wanted a more compact COGC/POPI management screen:

Before
image

After
image


purpleorangeazure:
Thank you so much for all of your feedback! I will work on improving all of my PRs with it in mind. I re-did this feature to use a button, as you suggested. I decided to give this button three states as I felt a simple lock/unlock might be annoying to use in the following cases:
 - If scroll lock prevented scrolling: you have the scroll locked, but you want to see another part of the tile without losing your saved position
 - If scroll lock did not prevent scrolling: the scroll lock button indicates that the scroll is saved, but you've scrolled from that position making it unclear which position is saved
 - If scrolling while saved just gave a save button: it's unclear if the scroll is saved or not at all

So, I made the state of saved-but-not-current-position a "go to saved scroll" button. A double-click unsaves. I considered making a right click immediately unsave/resave at the new position but I don't know how to effectively communicate that that is the behavior through the ui.

It was very difficult to pick out appropriate icons for the button states. I did not want to use colors, and my experiments with stacking a mouse icon and a star and lock icon to indicate state did not go well.

Fun features:

Tiles without scrollable content do not have the button
image

You can categorize your FLT buffer
image

So much information
image

Auto-maxing contribution sliders would even make _this_ a viable setup
image

purpleorangeazure:
Up to eight (maybe nine?) buffer controls is a little bit unwieldy, isn't it: thoughts on an overflow menu for them? As far as I know there's no analogous UI in PrUn, though, so it might be difficult to make it look/feel natural...

PR #138, opened 2025-12-28
Prunstats shortcut

purpleorangeazure:
image

Simple feature to add buttons to the USR and CO tiles that bring up XIT PRUNSTATS presets: One for products the user produces, and one for production volume history.

I feel that this is invaluable in getting to know the people you're interacting with by lowering the barrier to looking at what they produce.

Known issues:
 - Some users may not have data available in pmmg-products, making the buttons lead to a blank screen without any feedback. This might be more of an issue with pmmg-products itself.
 -  The CO shortcuts rely on the index of the 'Managing Director' field in the form. This is liable to break with updates. I tried using username matching through usersStore but it was unreliable.

It might be worth getting rid of the "hideOptions" parameter in the shortcuts, but personally I like the simplicity of it. This may just be because I am the one who chose the presets anyways.


purpleorangeazure:
Redid this a bit, I would really like to use the neutral button intent but it is basically unreadable.

image

This brings me to a larger issue; Is there any will to recolor the buttons so we have more contrast? Even the white-on-yellow can be almost unreadable sometimes.

image

Personally I think that #252525 for the text looks much, much better.

PR #139, opened 2025-12-28
Add slider features and fixes

purpleorangeazure:
image

Mainly, min/max buttons on both individual sliders and entire columns of them. Also, a few improvements/bug fixes that might be better off rolled into prun-bugs.

I have doubts about how well this (specifically, enabling/disabling) responds to external changes in the sliders but I'm not sure how to make it more resilient without a mutationobserver for every slider. I am also not sure if that would even be as bad a solution as it sounds.

I noticed that automatic class name registration in select-dom is not working. Or maybe I was just using it wrong. Either way, it almost drove me insane, as such things tend to do.

The min button is of questionable utility. Also, this makes sliders less wide, which can make granular changes more difficult. It might be better to put a number input on the left side instead of a min button.

Finally, I am aware that the mark-texts for the min and max also act as min/max buttons. I had no idea of this until I started digging around the code, and I doubt many other people noticed this either. Besides, the click targets are so small that the buttons make better UX anyways.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+- `contribution-bulk-controls`: Adds NONE/ALL buttons to contribution sections
+- `contribution-maxed`: Automatically maxes contribution sliders in CoGC and population upkeep tiles
 - `flt-flex-fuel`: Allows the fuel column layout to better use available space
 - `sysi-blue-negative-value`: Makes lower negative planet values blue instead of red
 
@@ -17,6 +19,8 @@
 
 - `prun-bugs`: Fix the dot / arrow in system info being left skewed
 - `prun-bugs`: Fix layout shift when selecting inventory grid items
+- `prun-bugs`: Fix slider dot stretching and cursor styles
+- `prun-bugs`: Disable POPID sliders that can't be filled due to full reserves
 - Fix financial data collection failure when the user has no warehouses
 
 ## 26.1.24


Razenpok:
Adding slider controls everywhere in the game sounds too bloat-y to me. You can read about my stance on adding new UI elements here:
https://github.com/refined-prun/refined-prun/pull/80#issuecomment-2759868324

If I recall correctly, the only places where bulk min/max can matter are the contribution sections (CoGC, POPI). These are the main places where players accidentally set the slider to the wrong position. The other places either don't see min/max values often (like SFC and MTRA), or are used so infrequently that we can ignore them for now (like HQ and SHYP)

So, let's make this feature simpler and cleaner
1. Max all sliders in all contribution sections by default
2. Add two buttons to the contribution section: "NONE" and "ALL" (or "MIN" and "MAX"). The former sets all sliders to 0, and the latter sets them to 100%. This one is very optional, as I don't see the need for any of these buttons. You can decide whether it is worth adding.

To match the contribution sections specifically, match the `C.Contribution.contribute` action bar - the one that contains the yellow "CONTRIBUTE" button. Once you match it, take a `.previousElementSibling`, check that it is a `<table>`, and then work with sliders in it.

By limiting the feature to the contribution sections, we can avoid accidentally breaking other features (like `XIT ACT MTRA`, I guess?). 

Please let me know if you disagree with my ideas so we can flesh out this feature. If you are going to proceed with the direction I proposed, rename this feature to `contribution-maxed`. And if you decide that the "NONE" and "ALL" buttons are a good idea, then put them into a separate `contribution-bulk-controls` feature.

I'm pretty sure people will let us know if they want HQ and SHYP slider maxing next, so it can be done in a future PR.

I will also leave a code review for this PR, including the code that will be removed, so you can start integrating the feedback into the new version.

image

src/features/basic/slider-improvements/slider-improvements.ts
```diff
@@ -0,0 +1,130 @@
...
+import { clickElement } from '@src/util';
+
+function onTileReady(tile: PrunTile) {
+  sliderMinMaxButtons(tile);
+  sliderTableBulkMinMax(tile);
+}
+
+function sliderTableBulkMinMax(tile: PrunTile) {
+  subscribe($$(tile.anchor, 'table'), table => {
+    const headerRow = _$(table, 'thead')?.firstElementChild;
```
Razenpok: Avoid using `_$`. Instead, make the function to `async` and use `await $` - it will give you a non-undefined element.

src/features/basic/slider-improvements/slider-improvements.ts
```diff
@@ -0,0 +1,130 @@
...
+    if (!headerRow) return;
+    const sliders = _$$(table, 'rc-slider');
+    if (sliders.length > 0) {
```
Razenpok: This `if` could be inverted to `sliders.length === 0` to reduce code nesting with an early `return`

src/features/basic/slider-improvements/slider-improvements.ts
```diff
@@ -0,0 +1,130 @@
...
+      const sliderColumns = new Map<number, HTMLElement[]>();
+      sliders.forEach(slider => {
```
Razenpok: Code style for this project: never use `.forEach`
```suggestion
      for (const slider of sliders) {
```

src/features/basic/slider-improvements/slider-improvements.ts
```diff
@@ -0,0 +1,130 @@
...
+        const cell = slider.closest('td');
+        const columnIndex = Array.prototype.slice.call(cell?.parentElement?.children).indexOf(cell);
```
Razenpok: Why do you call `.slice` via `Array.prototype.slice.call` instead of calling it directly?

src/features/basic/slider-improvements/slider-improvements.ts
```diff
@@ -0,0 +1,130 @@
...
+        };
+        const maxSliderColumn = () => {
+          sliders.forEach(slider => {
+            const markContainer = _$(slider, 'rc-slider-mark');
+            if (markContainer && !slider.classList.contains('rc-slider-disabled')) {
+              clickElement(markContainer.lastElementChild as HTMLElement);
+            }
+          });
+        };
+        const disabled = sliders.every(slider => slider.classList.contains('rc-slider-disabled'));
```
Razenpok: Code style for this project: if a lambda has only one parameter, call it `x`. Takes less time to think about a proper name, and reads as clearly as with a full name. Use full names when `x` results in an unclear code.
```suggestion
      const disabled = sliders.every(x => x.classList.contains('rc-slider-disabled'));
```

src/features/basic/slider-improvements/slider-improvements.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// ONLY FOR POPID(?)
+function disableInvalidSliders(tile: PrunTile) {
+  subscribe($$(tile.anchor, 'table'), table => {
+    subscribe($$(table, 'rc-slider'), slider => {
+      const sliderMaxMark = _$(slider, 'rc-slider-mark')?.lastElementChild;
+      if (!sliderMaxMark) return;
+      const sliderMax = parseFloat(sliderMaxMark.textContent);
+      const sliderValueMark = _$$(slider, 'rc-slider-mark-text-active').findLast(() => true); // the last active mark-text will be either the min/max or the current value
```
Razenpok: Code style for this project: put comments on a separate line, start with a capital letter, and end with a full stop.
```suggestion
      // The last active mark-text will be either the min/max or the current value.
      const sliderValueMark = _$$(slider, 'rc-slider-mark-text-active').findLast(() => true);
```

src/features/basic/slider-improvements/slider-improvements.ts
```diff
@@ -0,0 +1,130 @@
...
+      reactive({
+        onClick: maxSlider,
+        label: '>',
+        disabled,
+      }),
+    ).after(slider);
+  });
+}
+
+// ONLY FOR POPID(?)
```
Razenpok: This comment looks useless to me

src/features/basic/slider-improvements/SliderButton.vue
```diff
@@ -0,0 +1,12 @@
+<script setup lang="ts">
+defineProps<{ onClick?: () => void; label: string; disabled?: boolean }>();
+</script>
+
+<template>
+  <button
+    :class="` ${disabled ? C.Button.disabledInline : C.Button.primaryInline} ${disabled ? C.Button.disabled : C.Button.primary} ${C.Button.inline} ${C.Button.btn} `"
```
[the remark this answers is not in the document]
purpleorangeazure: Thank you, that first conditional syntax looks quite nice but I think in this case I would use the second since it properly shows the relationship between primary and disabled and I am more familiar with ternaries than Vue besides.

As for the last tip: Yes, I could and should have. I tried looking for generic components in the code but clearly I did not look hard enough.

PR #141, opened 2025-12-31
Tab Renaming

agm-114:
I created a feature that watches the selected the page and then updates the tab name if it changes.


src/features/basic/rename-tab.ts
```diff
@@ -0,0 +1,14 @@
+function updateTabTitle() {
+  const screenName = document
+    .querySelector(`.${C.ScreenControls.currentScreenName}`)
+    ?.textContent?.trim();
```
Razenpok: You can get the current screen name using this:
```suggestion
import { screensStore } from '@src/infrastructure/prun-api/data/screens';

  const screenName = screensStore.current.value?.name;
```

src/features/basic/rename-tab.ts
```diff
@@ -0,0 +1,14 @@
...
+  document.title = screenName ? `${screenName} - Prosperous Universe` : 'Prosperous Universe';
```
Razenpok: This feature doesn't account for notifications. It should include the notification count in parentheses before the tab name.

To get the notification count, use `alertsStore.all.value?.count ?? 0`

Then wrap all of this in `watchEffect` so the title updates automatically on both name and notification count updates.

```suggestion
  watchEffect(() => {
    const screenName = screensStore.current.value?.name;
    const notificationCount = alertsStore.all.value?.count ?? 0;
    document.title = // do your logic here;
  });
```

image

src/features/basic/rename-tab.ts
```diff
@@ -0,0 +1,14 @@
...
+}
+
+function init() {
+  // Listen for URL changes using the hashchange event
+  window.addEventListener('hashchange', updateTabTitle);
+  updateTabTitle();
```
Razenpok: hashchange won't be needed once you implement the changes above. Inline `updateTabTitle` into `init`.

Razenpok (changes requested):
Rename the feature to `browser-tab-name`

PR #146, opened 2026-01-14
Detailed ship inventory bar

agm-114:
image

Have the inventory bar for ships display a rough preview of their contents.


razenpok (changes requested):

YuLun-bili:
does this work with `SHPT`?

PR #148, opened 2026-01-14
Detailed Fleet Status

agm-114:
image


YuLun-bili:
Make it so that can toggle off showing status as an icon? I personally don't find the 'status icon' feature very helpful as status strings are max 4 characters in Chinese, typically 3 so it doesn't occupy much space.

And I changed remaining time to Chinese as well, since that's the one would be longer than English, if you get the string from vanilla `FLT` list rather than your own implementation. However that would make the bottom most case invalid as game just show the 2 most significant ones

*F12 edited & original
imageimage

en & zh-CN lookup if you want to test something
```
approaching			进近中
charging			充能中
waiting for decay	等待场衰
departing			离港中
floating			悬飘中
jumping				跃迁中
jumping gateway		跃门中
landing				着陆中
waiting for lock	等待对锁
taking off			起飞中
in transit			转移中
stationary			停泊

d / day		天
h / hour	小时
m / minute	分钟
```

PR #152, opened 2026-01-18
Add flt-flex-fuel

agm-114:
image
Allows the fuel bars to be displayed in a row if there is enough space.
This means stl & ftl fuel data will be vertically aligned & makes future optimisation of the FLT screen easer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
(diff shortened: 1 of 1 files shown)


ebitkov:
How about that, or is that too small? 

image

Basically merging them together to one bar.

erendrake (commented):
small nitpicks but overall its a cool option

PR #165, opened 2026-02-17
adding prod and wrk buttons for burn

### LINKED ISSUE #107: Add per-buffer toggle in XIT BURN to disable workforce consumption
Currently, XIT BURN consumes workforce resources uniformly, which can lead to excessive storage usage for workforce consumables compared to production inputs. A per-buffer toggle to disable workforce consumption would allow users to set different effective stock levels for workforce consumables and production inputs, optimizing storage and resource management. This feature would enhance flexibility in managing workforce consumption on a per-buffer basis.

Tracked in https://discord.com/channels/667551433503014924/1354424275670401096

erendrake:
Closes #107


razenpok:
Implemented in ec525bde

PR #195, opened 2026-04-03
Added XIT MATIO (material inputs+outputs)

richjenks:
**Summary**: Added columns to XIT BURN for production/workforce input quantity and production output quantity:

image

**Problem**: On bases which produce and consume the same material (e.g. RAT or H2O) it's unclear how much to keep on-site for local consumption vs how much is surplus to be sold. The current workaround is to open the plan in PRUNplanner, which adds friction and requires maintenance.

**Solution**: Show In and Out values separately, similar to Material I/O in PRUNplanner, so you know how much to keep (e.g. x times y days) and how much to sell.


Razenpok:
What if INV showed [current/required] on items with consumption instead? Like fuel tanks do:

image

This would solve your problem better than the proposed BURN change.

Relevant thread: https://discord.com/channels/667551433503014924/1310995731640877161/threads/1439652066921418852

richjenks:
@Razenpok good shout — that solves the problem with fewer elements and less mental arithmetic.

I've pushed this to the branch, lmk thoughts.

image

Razenpok:
I was talking about "let's add X/Y to INV buffer", not "let's add X/Y to Inv column in XIT BURN"

richjenks:
I think XIT BURN is the best place for this because it relates to material inputs/outputs rather than material management more generally.

Whereas doing it in INV:

image

1. Will get visually messy quite quickly
2. Creates an invisible and un-self-explanatory relationship between INV and XIT SET (whereas XIT BURN already uses the Resupply option in the Need column)
3. Splits your attention between INV and XIT BURN to answer related questions, as opposed to communicating everything in one buffer

Third options include making this a togglable setting (which is advised against in contributing docs) or a new buffer altogether that's essentially the rprun version of PRUNplanner's Material I/O table (but that's a heavier change while this is quite a simple improvement).

Can I ask what your objection is to doing it in XIT BURN?

erendrake:
Agreeing with @Razenpok — INV is the right home for this:
                                                                                                                                                                                                     
  1. Pattern already exists. Fuel tanks in INV use current / required. Extending that to consumables reuses vocabulary players already know instead of inventing a second "required" concept in a    
  different buffer.                                                                                                                                                                                  
  2. INV is closer to the moment of decision. Inventory decisions happen all over, but when you're actually transferring or selling goods, INV is what's in front of you — so the "keep vs sell?"
  answer is most useful there.                                                                                                                                                                       
                                                                                                                                               
                                                                                                                                                                                                     
  Separate correctness issue: requiredInvAmt uses two different formulas depending on whether needAmt > 0 (net vs gross consumption). Mixed producer/consumer bases — the motivating RAT/H2O case —  
  are exactly where those diverge, so the displayed meaning flips with inventory level. wouldnt it be confusing for that to change as you drag goods around?

richjenks:
The pattern exists but if you add metric suffixes (like fuel) it's imprecise and if you omit them (my previous comment) it's messy.

INV probably is closer to the decision UI (unless you're using MTRA) but UI bloat, inconsistency between materials with vs without consumption, numerators being higher than denominators, etc...feels like a solution requiring a bit more thought, whereas adding a single number to XIT BURN is a quick uncomplicated fix to a daily pain.

I read the thread and contributing.md but feels like there's context or principles I'm missing. What's the objection to solving in XIT BURN?

erendrake:
im not speaking for the project or @Razenpok on this but for me I am worried about XIT BURN being a bit of a design Christmas tree. Because its a very well used and familiar buffer its easy for everyone to hang their own features on it to the point that it gets impossible to understand when you open it for the first time. I think that the remit of each buffer is a bit hard to piece together from the current source code and that can lead everyone to think they know what its "for". 

In my opinion xit burn is more or less complete. I have a PR to add more filter buttons which might get abandoned as im not sure its worth it anymore.

richjenks:
> I think that the remit of each buffer is a bit hard to piece together from the current source code and that can lead everyone to think they know what its "for".

12/10 product thinking. If XIT BURN is for burn rates and how long materials on the base will last, it already does that well, and the problem of general material i/o data is a separate one.

That said, I still don't think INV is the right place for that for various reasons listed above. Also, if you have 0 of a material you need, it's not in INV at all. Also, it wouldn't make sense in SHPI, despite that potentially being a useful place for it. In short, I don't believe that's the right solution for the problem.

So instead I'm proposing a new XIT MATIO buffer altogether with the separate goal of providing material in and out data, similar to prunplanner, which will always have correct info and avoids having to switch to prunplanner every time you go shopping.

I'm more than happy to pivot this PR that way if people agree :)

raylu:
I was thinking of implementing exactly this (splitting net burn into consumption/production. very glad richjenks already implemented it!

> 2. INV is closer to the moment of decision. Inventory decisions happen all over, but when you're actually transferring or selling goods, INV is what's in front of you — so the "keep vs sell?" answer is most useful there.

in fact, I wanted this feature specifically to decide what to keep vs sell. are you imagining that
- in a base inventory, the numbers would show that base's consumption/production? how would you know how much of the item you have at the base (you have 10 RAT, you consume 5 a day, you produce 12 a day; what should it show)?
- in a CX warehouse inventory, the numbers would show your overall consumption/production? maybe the consumption for bases that are nearer to that CX than other CXes?

what would it do for items in a ship inventory? planet warehouse? other storages?
is there a way to see the combined consumption/production for a subset of bases?
splitting net burn into consumption/production in XIT ACT solves all these problems
I wouldn't mind XIT MATIO either, though

EDIT: I think razenpok/erendrake are saying that INV would show current/required rather than consumption/production. I don't think that helps for selling things: you generally sell things that you're producing more of than you're consuming, so the required is 0

erendrake:
@richjenks i like XIT MATIO. Once it's in people can try it and we'll see whether it displaces BURN for this use case or sits alongside it. Either outcome is fine, and it keeps BURN focused on burn rates.

richjenks:
Pivoted so BURN remains as-is (material burn rate) with new features added to MATIO (material in/out). Feels right to split focus rather than overlap, makes both more useful and less noisy.

image

Fwiw I'd have both on my bases screen because they're useful for different reasons.

raylu:
I've been using this branch and I just have a MATIO right next to my BURN. I'd still basically like to see the in/out separately in BURN or the days/remaining and color coding on MATIO

lumivient:
Fixed by https://github.com/refined-prun/refined-prun/commit/ec525bdecd21364e7f65fe2382757e550b5af562

PR #196, opened 2026-04-05
prun-bugs: Prevent text selection when moving items by dragging

richjenks:
Fixes an annoyance on Chrome for Mac (and likely others) where draggable elements can allow inadvertent user selection, e.g. when moving materials between INV/SHPI buffers, half your screen gets selected.

inv

Implemented in a way that should work for all draggable elements, not just materials.


src/features/basic/prun-bugs.ts
```diff
@@ -66,10 +66,37 @@ function fixSliders() {
...
+
+  document.addEventListener('dragstart', e => {
+    const target = e.target as Element;
+    if (target.closest('[draggable="true"]') === null) {
+      return;
+    }
+    document.body.classList.add(noSelectClass);
+  });
+
+  window.addEventListener('pointerup', queueReset);
```
erendrake: pointerup isnt always guaranteed, esp if the user drags outside the browser window. it seems like listening to dragend would be wise to avoid the user getting stuck in a not select state

erendrake (commented):
I think we are missing an event

richjenks:
Thanks @erendrake , good catch, fixed!

PR #199, opened 2026-04-09
XIT ACT: Prevent manually entered quantity/price overwrite by order book changes in CX Buy

raylu:
CXOB updates cause a recalculation of the price,
which unexpectedly reverts manually-entered quantity/price.

PR #200, opened 2026-04-09
XIT ACT CX Sell

raylu:
Add a new XIT ACT action type that sells things instead of buying them.

The only sensible material group to use for this is probably manual.
I plan to write some code to look at the inventory of ships that just landed at CX
and figure out what to sell vs keep, then generate a CX Sell action.

### DIFF
--- a/src/features/XIT/ACT/action-steps/CXPO_SELL.ts
+++ b/src/features/XIT/ACT/action-steps/CXPO_SELL.ts
@@ -1,0 +1,181 @@
+import { act } from '@src/features/XIT/ACT/act-registry';
+import { fixed0, fixed02 } from '@src/utils/format';
+import { changeInputValue, clickElement } from '@src/util';
+import { fillAmount } from '@src/features/XIT/ACT/actions/cx-sell/utils';
+import { storagesStore } from '@src/infrastructure/prun-api/data/storage';
+import { exchangesStore } from '@src/infrastructure/prun-api/data/exchanges';
+import { warehousesStore } from '@src/infrastructure/prun-api/data/warehouses';
+import { watchWhile } from '@src/utils/watch';
+import { materialsStore } from '@src/infrastructure/prun-api/data/materials';
+import { watchEffect } from 'vue';
+import { AssertFn } from '@src/features/XIT/ACT/shared-types';
+
+interface Data {
+  exchange: string;
+  ticker: string;
+  amount: number;
+  priceLimit: number;
+  sellPartial: boolean;
+  allowUnfilled: boolean;
+}
+
+export const CXPO_SELL = act.addActionStep<Data>({
+  type: 'CXPO_SELL',
+  preProcessData: data => ({ ...data, ticker: data.ticker.toUpperCase() }),
+  description: data => {
+    const { ticker, exchange } = data;
+    const cxTicker = `${ticker}.${exchange}`;
+    const filled = fillAmount(cxTicker, data.amount, data.priceLimit);
+    const amount = filled?.amount ?? data.amount;
+    const priceLimit = filled?.priceLimit ?? data.priceLimit;
+    const allowUnfilled = data.allowUnfilled ?? false;
+    const willFillCompletely = filled && filled.amount === data.amount;
+
+    if (!willFillCompletely && allowUnfilled) {
+      let description = `Ask for ${fixed0(data.amount)} ${ticker} on ${exchange}`;
+      if (priceLimit > 0) {
+        description += ` at price ${fixed02(data.priceLimit)}`;
+      }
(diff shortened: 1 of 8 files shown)


erendrake:
Any time we are adding new features its important to know what is getting added and why and what is in and out of scope before we start to tear apart the code. PR body is very helpful

src/features/XIT/ACT/action-steps/CXPO_SELL.ts
```diff
@@ -0,0 +1,181 @@
...
+    const amount = filled?.amount ?? data.amount;
+    const priceLimit = filled?.priceLimit ?? data.priceLimit;
+    const allowUnfilled = data.allowUnfilled ?? false;
+    const willFillCompletely = filled && filled.amount === data.amount;
+
+    if (!willFillCompletely && allowUnfilled) {
+      let description = `Ask for ${fixed0(data.amount)} ${ticker} on ${exchange}`;
+      if (priceLimit > 0) {
+        description += ` at price ${fixed02(data.priceLimit)}`;
+      }
```
erendrake, line 38: Symmetry break vs `CXPO_BUY.ts`: the buy-side `!willFillCompletely && allowUnfilled` branch appends a `(X total cost)` line after the price. Sell should mirror this with a `(X total value)` line — `data.amount * data.priceLimit` for consistency with how buy presents the same information. Suggested:

```ts
if (priceLimit > 0) {
  description += ` at price ${fixed02(data.priceLimit)}`;
  description += ` (${fixed0(data.amount * data.priceLimit)} total value)`;
}
```

src/features/XIT/ACT/action-steps/CXPO_SELL.ts
```diff
@@ -0,0 +1,181 @@
...
+        changeInputValue(quantityInput, filled.amount.toString());
+        changeInputValue(priceInput, fixed02(filled.priceLimit));
+      }
+
+      // Cache description before clicking the sell button because
+      // order book data will change after that.
+      ctx.cacheDescription();
+    });
+
+    await waitAct();
```
erendrake, line 168: Cross-PR note: PR #199 adds an input-event guard to `CXPO_BUY.ts` so CXOB updates don't overwrite a user's manually-edited quantity/price while `waitAct()` is pending. `CXPO_SELL.ts` has the same bug — it's structured identically to pre-#199 `CXPO_BUY.ts`. Whichever of #199 or #200 lands second should port the fix so both files stay aligned.

src/features/XIT/ACT/actions/cx-sell/Edit.vue
```diff
@@ -0,0 +1,79 @@
...
+  </Active>
+  <Active label="Exchange">
+    <SelectInput v-model="exchange" :options="exchanges" />
+  </Active>
+  <Commands label="Minimum Prices">
+    <PrunButton primary @click="onEditPriceLimitsClick">EDIT</PrunButton>
+  </Commands>
+  <Active
+    label="Sell Partial"
+    tooltip="Whether the action will be taken if there is not enough demand on the CX.">
```
erendrake, line 120: UX label mismatch: this form shows `Minimum Prices`, but clicking EDIT opens the reused `EditPriceLimits.vue` from `cx-buy/`, which renders `Edit Price Limits` as the section header and `Price Limit #N` for each row. Users will see two different names for the same concept in a single flow.

Options: rename this label to `Price Limits` to match the overlay, or parameterize `EditPriceLimits` with a label prop so sell can pass `Minimum Price #N`.

raylu:
wrote some words in the PR desc

src/features/XIT/ACT/action-steps/CXPO_SELL.ts
raylu: yeah exactly! my plan was to edit whichever didn't land first

src/features/XIT/ACT/actions/cx-sell/Edit.vue
raylu: wow, great catch

src/features/XIT/ACT/action-steps/CXPO_SELL.ts
raylu: I brought over the change in that PR to this one. I need it even more here since I tweak both quantity and prices a lot when I sell

raylu:
added some code to configure which inventory the selling happens from

when buying, the "storage location" select doesn't really matter (things seem to always go into your warehouse if you have space, no matter what you select) so CXPO BUY treats the CX warehouse as the buy target always

when selling, my workflow is to sell the stuff straight from the ship. everything I don't sell I plan to keep so I unload afterwards

PR #201, opened 2026-04-09
POPID Visible Contribution in Bar

purpleorangeazure:
image

Hoping to make contribution-maxed more user-friendly when applied to POPID by distinguishing the reserve stock from the not-yet-contributed stock. This should ideally apply in many more contribution bars. 

The way that PrUn counts the reserve seems a bit strange; when contributing to an upkeep that has no currently active consumption, the reserve bar does not subtract the consumption from the to-be-contributed amount. I have not yet been able to test whether this is properly calculated after contribution. This discrepancy makes it difficult to determine whether the consumption bar should be marked as to-be-contributed or if it was previously filled. 

I think that the yet-to-be-contributed part should be a different color, but I could not think of one so I settled on a striped design akin to the loader pattern.


purpleorangeazure:
image
I had a realization about the Next Consumption bar (it is not a 'real' reserve) and implemented the stacked progress bar in that cell as well.

PR #202, opened 2026-04-14
CONTD: Confirm dialog before applying template with unselected commodity

erendrake:
## Summary
When creating a contract draft, typing a commodity ticker without pressing Enter caused PrUn to silently default the field to rations. This adds a PrUn-styled confirmation overlay that lists the affected commodities with **Cancel** / **Apply anyway** options.

Detection tracks Enter keypress and suggestion-click events on each MaterialSelector to record committed values. Falls back to `materialsStore.getByName` for pre-loaded forms.

## Test plan
- [ ] Open CONTD, choose BUY, type a ticker (e.g. `O`), do **not** press Enter, click "apply template" → confirm dialog appears
- [ ] Same flow but press Enter on the ticker → no dialog
- [ ] Add a second commodity, leave one bad and one good → dialog lists only the bad one with correct index
- [ ] Click "Apply anyway" → template applies with rations (verifying we don't double-block)
- [ ] Click "Cancel" → no template applied, can fix and retry

### DIFF
--- a/src/features/basic/contd-template-confirm-uncommitted-material.ts
+++ b/src/features/basic/contd-template-confirm-uncommitted-material.ts
@@ -1,0 +1,95 @@
+import { materialsStore } from '@src/infrastructure/prun-api/data/materials';
+import { showConfirmationOverlay } from '@src/infrastructure/prun-ui/tile-overlay';
+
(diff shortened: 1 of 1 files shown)


erendrake:
Closing to rework as a stacked PR with gh-stack. Will reopen as the bottom of the stack.

Issue #170, opened 2026-02-28
XIT BURN: Inaccurate days remaining for fractional burn rates (< 1 unit/day)


erendrake:
## Description

When workforce consumption is less than 1 unit per day (e.g., 0.08 SCN/day for 80 Technicians), the BURN days remaining calculation can be wildly inaccurate because the game API reports inventory amounts as integers.

## Scenario

- 80 Technicians consuming 0.08 units/day of SCN
- Inventory shows **1 unit** remaining
- BURN displays **~12 days** remaining (`floor(1 / 0.08) = 12`)
- In reality, that "1 unit" could be anywhere from 0.01 to 1.0 internally
- The material was fully consumed overnight — actual remaining time was **< 1 day**, not 12

## Confirmed: Game API only provides integer inventory

Verified via WebSocket inspection (`STORAGE_CHANGE` message). The `quantity.amount` field is always an integer. The `value.amount` field is the monetary value (currency), not the material count. Example:

```json
{
  "quantity": {
    "value": { "amount": 0.33, "currency": "CIS" },
    "material": { "ticker": "C", ... },
    "amount": 292
  }
}
```

All inventory amounts observed are integers (292, 92, 1300, 28, etc.). The game rounds internally before reporting, so when inventory shows `1`, the actual internal amount could be anywhere from 0.01 to 1.0.

## Root Cause

`src/core/burn.ts` divides integer inventory by the fractional daily burn rate:

```typescript
daysLeft = Math.floor(-materialBurn.inventory / materialBurn.dailyAmount);
```

When `inventory = 1` and `dailyAmount = -0.08`, this gives `floor(12.5) = 12` days. But the real internal amount may be nearly zero.

## Impact

- BURN display shows misleading days remaining for low-consumption materials
- Resupply calculations (`src/features/XIT/ACT/material-groups/resupply/resupply.ts`) use the inflated `daysLeft`, causing shipments to arrive too late
- Users experience unexpected stockouts despite BURN showing days of supply remaining

## Possible Solutions

1. **Pessimistic estimate for the last unit**: When burn rate is fractional (< 1/day), subtract 1 from inventory before calculating days. Shows 0 days instead of 12 for the scenario above.
2. **Range display**: Show "0–12 days" to communicate the uncertainty when on the last consumed unit.
3. **Warning indicator**: Flag materials where `inventory < ceil(1 / burnRate)` as unreliable, since the game's integer rounding makes the estimate meaningless at that threshold.
[...]

PR #204, opened 2026-04-15
Add ACT button to BURN for one-click planet resupply

jackinabox86:
I vibe-coded a feature to add an ACT button to the XIT BURN buffer that opens a preconfigured action package to resupply that planet.  It's preconfigured to use the number of days the user has set for resupply in XIT SET, but is user-configurable.  It has a configurable CX Buy action and then a configurable MTRA action.  

Beyond convenience, I think it is most helpful for new players who don't know how to set up an action package and properly set up material groups and cx buy/mtra actions themselves.  

Given I am not an experienced programmer, I put this here more as inspiration if someone with a higher-skill level wants to implement it.  I did run the agents.md reviews and believe in good-faith that the repo comports with those and the game's underlying mod restrictions.  

https://github.com/user-attachments/assets/789b8676-c471-4fbe-93b1-f8900024a705

  Adds an ACT button next to BS/INV in XIT BURN planet rows. Clicking it                                                                                                                      
  opens XIT BURNACT <naturalId> which builds a transient resupply action
  package (Resupply group → CX Buy → MTRA) and shows the execution UI.  
                                                                                                                                                                                              
  - Resupply days and planet are pre-filled from the clicked row and the
    XIT SET burn.resupply setting, both overridable in the configure window                                                                                                                   
  - Adds configure-on-execution support for the resupply days field        
  - Adds configure-on-execution support for the CX Buy exchange field                                                                                                                         
  - New XIT BURNACT command renders ExecuteActionPackage with a transient
    (non-persisted) ActionPackageData — no saved package required

### DIFF
--- a/src/features/XIT/ACT/actions/cx-buy/Configure.vue
+++ b/src/features/XIT/ACT/actions/cx-buy/Configure.vue
@@ -7,9 +7,7 @@ const { config } = defineProps<{ data: UserData.ActionData; config: Config }>();
 
 const exchanges = ['AI1', 'CI1', 'IC1', 'NC1', 'CI2', 'NC2'];
 
-if (!config.exchange) {
-  config.exchange = exchanges[0];
-}
+config.exchange ??= exchanges[0];
 </script>
 
 <template>
--- a/src/features/XIT/ACT/actions/cx-buy/Configure.vue
+++ b/src/features/XIT/ACT/actions/cx-buy/Configure.vue
@@ -1,0 +1,21 @@
+<script setup lang="ts">
+import Active from '@src/components/forms/Active.vue';
+import SelectInput from '@src/components/forms/SelectInput.vue';
+import { Config } from '@src/features/XIT/ACT/actions/cx-buy/config';
+
+const { config } = defineProps<{ data: UserData.ActionData; config: Config }>();
+
+const exchanges = ['AI1', 'CI1', 'IC1', 'NC1', 'CI2', 'NC2'];
+
+if (!config.exchange) {
+  config.exchange = exchanges[0];
+}
+</script>
+
+<template>
+  <form>
+    <Active label="Exchange">
+      <SelectInput v-model="config.exchange" :options="exchanges" />
+    </Active>
+  </form>
+</template>
--- a/src/features/XIT/ACT/actions/cx-buy/Edit.vue
+++ b/src/features/XIT/ACT/actions/cx-buy/Edit.vue
@@ -7,6 +7,7 @@ import RadioItem from '@src/components/forms/RadioItem.vue';
 import { showTileOverlay } from '@src/infrastructure/prun-ui/tile-overlay';
 import EditPriceLimits from '@src/features/XIT/ACT/actions/cx-buy/EditPriceLimits.vue';
 import { materialsStore } from '@src/infrastructure/prun-api/data/materials';
+import { configurableValue } from '@src/features/XIT/ACT/shared-types';
 
 const { action, pkg } = defineProps<{
   action: UserData.ActionData;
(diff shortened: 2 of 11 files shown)


erendrake:
I am not speaking for @Razenpok or the project when i say this but i think this doesnt belong in XIT BURN. i had the same commit on #195 and I think a dedicated buffer with a grid of planets and saved actions would be really powerful. See the discussion over in #195

jackinabox86:
I had followed #195 and was disappointed at the outcome.  I'm not wedded to the IO implementation there (perhaps one could just parenthetical in red the burn of a consumed output), but locking BURN in statis would be unfortunate.  

Of course I'm in love with my own idea here, but I think this is a very natural upgrade for XIT Burn.  People monitor it, likely more than nearly any other buffer, and when it gets low, they generally use XIT ACT to address the situation.  

The current flow from noticing a low number to executing an action package is roundabout.  One has to open XIT Act, then either select a particular pre-configured planet action package or a general one that needs several configuration selections.  If users have already set a default resupply period, why not shorten that start-up time dramatically?  

This is more acute for new players who have very little guidance on how to create a new action package.  They have to create it, then learn about material groups, hope to figure out that the CX buy action needs to be added before the MTRA (an abbreviation they almost certainly don't know) action, etc.  Something like this allows a new player to add the extension and easily send a resupply ship with virtually no learning curve.

I'll defer to Razenpok, of course.  I do appreciate that one doesn't want BURN to become the one buffer that rules them all, but I do think this natural linking of BURN to ACT would be quite popular and doesn't detract from the simple presentation that BURN currently has; it operationalizes it.

Assuming something like this isn't implemented, I do hope there is a willingness to move the number of days for resupply from the material group setting to the configuration buffer.  This is what I currently have and it is unwieldy to have to set it correctly deep in the menus and then hope that the presets match future needs to avoid returning to those settings:

image

erendrake:
Appreciate the thoughtful response. Some specific feedback:

**On BURN as the home for this:** I don't think we should add horizontal space to BURN — not an ACT button, not a link-out, nothing. BURN's row layout is already tight (BS, INV, and the data columns), and every button added competes for space that matters more on narrow tiles. If there's a resupply workflow to build, it should be discoverable from its own buffer, not bolted onto BURN's planet rows. I don't share you opinion that something so useful that i have it on most of my screens not changing is sad. 

**On "configure days on execution":** It defaults to the XIT SET resupply value which is reasonable. No strong objection since it's opt-in.

**On the Need formula change:** `resupply.ts:85-86` replaces the existing formula with a different calculation that affects ALL Resupply material groups, not just BURNACT.

Old: `Math.ceil((matBurn.daysLeft - days) * matBurn.dailyAmount)`
New: `Math.max(0, Math.ceil(days * -matBurn.dailyAmount - matBurn.inventory + 1))`

These are not equivalent. Concrete example — RAT with `dailyAmount = -10`, `inventory = 150`, target 15 days:
- Old formula: `Math.ceil((15 - 15) * -10)` = **0** (already stocked, no delivery needed)
- New formula: `Math.max(0, Math.ceil(150 - 150 + 1))` = **1** (delivers an extra unit for no reason)

The `+1` always rounds up by one extra unit. A base exactly at target gets 1 unnecessary unit per material. A base needing 100 gets 101. This silently changes quantities for every player's saved Resupply packages. Needs to be reverted or split out with the `+1` justified.

P.S. are you addressing this issue we have been tracking https://github.com/refined-prun/refined-prun/issues/170 with this +1?

Deferring to @Razenpok on the final call.

jackinabox86:
Yes, I am tracking #170 on resupply calculation.  I've now commented there that the solution in #188 or my own is a very important change as we wait, likely months, for Molp to address the fractional issue.  Addressing that was a huge QOL change for me, as I assume Lilbit found because it made XIT BURN usable by preventing inaccurate 0s.  Including my fix here was unintentional though as it is out of scope of the discussion over potentially connecting BURN to ACT.  

On BURN itself, I think the miniscule number of pixels would be unlikely to impact the vast majority of users and would be a worthy sacrifice to significantly improve user flow.  I have separately followed your PR on XIT STO, which I think is cool and I hope is implemented, but same as one doesn't want BURN to explode in size, several new XIT ____s are likely to go unused and be lost in the shuffle.  If Razenpok/the community want to reject this, that is fair enough, but adding a third buffer to go with XIT BURN and XIT ACT would be an unwise detour; just polish XIT ACT at that point.

agm-114:
I'd say 50% of my responses to a yellow or red burn indicator involve xit act. 
My guess is that is below the average for vet players. 
However adding a act button bloats the menu for players not using act. 
My suggestion would be have the button enabled by default, but have a hidden settings menu a la xit flt for people who need a more compact layout. 
You could even throw in an option to hide the button column entirely.
Same for the plus +1 on consumables.

razenpok:
Merged in 26.9.15

PR #206, opened 2026-04-15
Add XIT STO

erendrake:
## Summary

Adds `XIT STO` (alias `XIT STORAGE`) — a per-base storage-analysis buffer that answers two questions at a glance:

1. **When will storage fill up?** — days-until-full at current net production rate.
2. **Will the next resupply fit?** — visual comparison of current inventory vs projected inventory after a full rotation (ship out produced goods + deliver consumables up to their XIT BURN \`Need\` amount).

Each base row shows:

- **Planet** — with a left-edge colored stripe (red / yellow / green) signaling how tight the resupply fit is.
- **Days Till Full** — single number; tooltip names the binding dimension (weight or volume).
- **Current Fill** — a `CargoBar` (category-colored stacked bar) showing what's in storage right now.
- **After Resupply** — a second `CargoBar` showing projected inventory post-rotation. If the resupply plan exceeds capacity, the bar enters a loud \"hazard mode\": muted category colors, a full-bar yellow hazard overlay, and a black-and-yellow striped overflow segment showing exactly how much over capacity the plan runs.
- **CMD** — BS / INV buttons.

Expanding a row reveals a multi-panel detail view (Fill Summary / Days of Supplies That Fit / Shipping Out / Adding / Visitation Frequency / Days Till Full) intended for triage.

### Collapsed view

image

### Expanded row

image

## Notes

> ⚠️ **Cherry-picks from #172** — This PR includes \`src/components/CargoBar.vue\` borrowed from @agm-114's [\`XIT FLT\` PR (#172)](https://github.com/refined-prun/refined-prun/pull/172). Minor generalizations were applied so the same component works for both ship cargo and base storage:
> - Takes \`store: Store\` directly instead of \`shipId\` (callers resolve the store).
> - Click handler is an optional prop (FLT passes a \`SHPI\` opener, STO omits).
> - Added \`disableMiniMode\` prop and an overflow-segment rendering mode for projections that exceed capacity.
>
> When #172 merges, one of us will need to reconcile the CargoBar location. Happy to rebase whichever way lands second.

## New helpers in \`core/\`

- \`computeNeed(materialBurn, resupplyDays)\` moved from \`BURN/utils.ts\` to \`core/burn.ts\` so it's shared between BURN and STO.
- \`BaseStorageAnalysis\` + \`getBaseStorageAnalysis\` + \`buildProjectedStore\` in \`core/storage-analysis.ts\`. The analysis is a \`computed\` map keyed by siteId (same reactive pattern as \`core/burn.ts\`). \`buildProjectedStore\` returns a synthetic \`Store\` with produced goods removed and consumables topped up to \`Need\`, which the CargoBar renders directly.

## Test plan

- [ ] \`XIT STO\` opens and lists all bases
- [ ] \`XIT STO Gibson\` / \`XIT STO JS-299b\` open with that base pre-expanded
- [ ] A base that can't fit a full resupply shows hazard styling on the After Resupply bar
- [ ] Hover CargoBar segments → instant PrUn-style tooltip with category + weight/volume
- [ ] Hover column headers → tooltips explain the metric
- [ ] Left-edge row stripe is red / yellow / green in line with projected fill ratio
- [ ] Expand a base → detail panels render (Fill Summary, Days of Supplies, Shipping Out, Adding, Visitation Frequency, Days Till Full)
- [ ] XIT BURN still shows the same Need values as before (regression check on the \`computeNeed\` move)
- [ ] Sort order: bases closest to filling first; bases that aren't filling (\`∞\`) sorted alphabetically at the bottom


agm-114:
My live reaction as a serial cookie licker.
tenor
This is way better than what I had planned : P though.

erendrake:
@agm-114 i am standing on the shoulders of your awesome PR and cargobar design. I a leaving this as draft until yours gets in because I think you deserve full credit for your part :)

agm-114:
image
I think by shuffling around the tooltip info a bit we might be able to compact it some.
Essentially compact stuff / fixed size stuff in the first row, expanding stuff in the second.

erendrake:
@agm-114 thanks and i agree that this panel is a bit of a mess. I kinda threw everything and the kitchen sink into that expando and figured we could sort it out in here. ill give this a try and likely collapse some of the lower data panels

erendrake:
@agm-114 here is a revision for you, i have added a little more detail 

image

i have tried several sizes of this panel and i think the expando needs a flow that will make rows of these panels as needed. 

here is a full width that has it all in one row 

image

and here is a much tighter one that has to flow down a few rows 

image


and if people dont need the extra content they can just have the numbers in a small buffer 

image

i almost didnt include supply days as it wont change until you modify your base in some way, either production or buildings?

agm-114:
Looks solid.

agm-114:
This is a bit of a tough one and a huge tangent. Idk if it fits in this pr or one with xit burn. 
But there's no indicator which planets have STL connections to the CX 
with the size of the gateway network this makes remembering when to use stl ships a bit tricky. 
Some sort of indicator of these systems would be good I think.

jackinabox86:
Can I ask why you brought back the header stripe in XIT BURN?

erendrake:
@jackinabox86 they will be removed, they were a test and them still being in is one of the reasons this is still in draft, thank you for the callout.

jackinabox86:
Just wanted to say I'm really enjoying the feature.  I won't suggest my implementation for a PR as the button violates the Refined PrUn style standards and it's your feature to expand, but I thought you might like to see someone is using it on their fork. 

https://github.com/user-attachments/assets/31c29469-a49f-49e4-b2d6-5dabfdf4254e

jackinabox86:
Now that there's a warehouse button in base inventory buffers, maybe you can add one to STO.  

Yes, I know; I'm always suggesting buttons.

PR #207, opened 2026-04-16
MTRA config: filter destination list based on selected origin

jackinabox86:
A QOL improvement that has been working well for me.  Filters the destination field in MTRA XIT ACT buffers to possible destinations when both origin and destination are set to "Configure on Execution".  Selecting an origin now filters the destination dropdown to same-location storages only.  Avoids extremely long lists of destinations to scroll through. 

https://github.com/user-attachments/assets/959eafe8-c291-47e4-a8d2-8b84dbdf0330


razenpok:
Cannot push changes to your branch because idk. I've cherry picked your changes manually in 10d4198d

image

PR #208, opened 2026-04-17
Add PROD button to XIT BURN for native PROD buffer

Thecat8413:
Adds a button to the XIT BURN command that opens the relevant base's native PROD menu, similar to BS or INV. I'm aware of the XIT PROD menu, but I personally prefer having the native PROD menu for when I want to view current production status, as the UI just works a bit better for me when quickly looking it over. 

image

Adding this button to XIT BURN is more of a QOL than anything else, as I am currently using XIT BURN as a central hub for all my bases on one screen. 

If needed, I can expand the change to the XIT PROD buffer as well, but I figured that was a bit redundant. 

There is also a PROD button on the overall; it should be hooked up properly to the native PROD menu.

### DIFF
--- a/src/core/burn.ts
+++ b/src/core/burn.ts
@@ -23,6 +23,7 @@ export interface BurnValues {
 }
 
 export interface PlanetBurn {
+  prodSiteId: string;
   storeId: string;
   planetName: string;
   naturalId: string;
@@ -48,6 +49,7 @@ const burnBySiteId = computed(() => {
         }
 
         return {
+          prodSiteId: id,
           storeId: storage?.[0]?.id,
           planetName: getEntityNameFromAddress(site.address),
           naturalId: getEntityNaturalIdFromAddress(site.address),
--- a/src/features/XIT/BURN/BURN.vue
+++ b/src/features/XIT/BURN/BURN.vue
@@ -124,7 +124,13 @@ const planetBurn = computed(() => {
     }
   }
 
-  const overallSection = { burn: overallBurn, planetName: 'Overall', naturalId: '', storeId: '' };
+  const overallSection = {
+    burn: overallBurn,
+    planetName: 'Overall',
+    naturalId: '',
+    prodSiteId: '',
+    storeId: '',
+  };
   if (queryResult.value.overallOnly) {
     return [overallSection];
   }
--- a/src/features/XIT/BURN/PlanetHeader.vue
+++ b/src/features/XIT/BURN/PlanetHeader.vue
@@ -30,6 +30,9 @@ const days = computed(() => countDays(burn.burn));
         <PrunButton dark inline @click="showBuffer(`INV ${burn.storeId.substring(0, 8)}`)">
           INV
         </PrunButton>
+        <PrunButton dark inline @click="showBuffer(`PROD ${burn.prodSiteId.substring(0, 8)}`)">
+          PROD
+        </PrunButton>
       </div>
     </td>
   </tr>
--- a/src/features/XIT/BURN/PlanetHeader.vue
+++ b/src/features/XIT/BURN/PlanetHeader.vue
@@ -4,6 +4,7 @@ import { showBuffer } from '@src/infrastructure/prun-ui/buffers';
 import PrunButton from '@src/components/PrunButton.vue';
 import { PlanetBurn } from '@src/core/burn';
 import { countDays } from '@src/features/XIT/BURN/utils';
+import { useXitParameters } from '@src/hooks/use-xit-parameters';
 
 const { burn } = defineProps<{
   burn: PlanetBurn;
@@ -11,6 +12,9 @@ const { burn } = defineProps<{
   minimized?: boolean;
   onClick: () => void;
 }>();
+
+const parameters = useXitParameters();
+const showProdButton = computed(() => parameters.length === 0);
 
 const days = computed(() => countDays(burn.burn));
 </script>
@@ -30,7 +34,11 @@ const days = computed(() => countDays(burn.burn));
         <PrunButton dark inline @click="showBuffer(`INV ${burn.storeId.substring(0, 8)}`)">
           INV
         </PrunButton>
-        <PrunButton dark inline @click="showBuffer(`PROD ${burn.prodSiteId.substring(0, 8)}`)">
+        <PrunButton
+          v-if="showProdButton"
+          dark
+          inline
+          @click="showBuffer(`PROD ${burn.prodSiteId.substring(0, 8)}`)">
           PROD
         </PrunButton>
       </div>


erendrake:
There's been a pattern of PRs adding buttons to XIT BURN:

- #165 — PROD/WRK toggles (mine — I'm part of the problem)
- #135 — consumption display for input shortages
- #204 — ACT button for resupply
- #208 — PROD button (this PR)

BURN's value is density. Players tile it 8+ across their screen. Every button added to the planet row costs horizontal space on every row for every user. BS and INV are already tight.

For PROD specifically: the BS button already opens the base screen, which has PROD in it. This saves one click at the cost of permanent row width.

I think we need to stop adding to BURN and route new functionality through dedicated buffers or context menus instead.

Thecat8413:
I can definitely understand the desire to prevent feature creep.

For the horizontal spacing issue, I think a decent middle ground would be to hide the button when the tile is base specific. Adding a PROD button was primarily for selecting the specific PROD window for a base when in a list of bases. For situations where a player is tiling XIT BURN I would assume that each tile is for a specific, unique base rather than the default XIT BURN buffer. 

I use the default window as the core for a single "bases" page that handles all of my bases rather than having base specific screens. For me at least, the only two things I regularly access that are not on that screen is the PROD buffers and LMs. There is a native default PROD window but it takes up a lot of space, so I don't like using it. As for LMs, I'm working on a new buffer that would handle that, but that's irrelevant to this PR.

I can alter the PR to make the button only appear when in the default buffer, just not this weekend.

agm-114:
My thought is add the prod button, but remove inv & bs. 
The base name can be a clickable link that takes you to bs.
You can get to inv through the link. 

I think if you really want an INV button having it disabled by default, but having a setting to enable it like xit fleet would be good.

Thecat8413:
> My thought is add the prod button, but remove inv & bs. The base name can be a clickable link that takes you to bs. You can get to inv through the link.
> 
> I think if you really want an INV button having it disabled by default, but having a setting to enable it like xit fleet would be good.

Clicking the name of the base (when in default mode) currently expands the base info for XIT BURN. That's core functionality, so changing that doesn't really work in my mind. 

Removing INV has the same issue as not adding PROD; it's a common action that a player would want easy access to. If anything, most users likely would want INV over PROD; most bases are "set-and-forget" for production, but inventory management is always needed. 

I'll take a look at the settings logic as well as hiding the PROD button when in non-default mode.

Thecat8413:
@erendrake Pushed new commit 69a85cb, which hides the PROD button when XIT BURN is given parameters. This avoids the issue with affecting XIT BURN tiling. Would you like me to add a default setting that hides the button completely as well, or does this address your concerns?

jackinabox86:
I'm guilty of doing the added button to XIT BURN, so take my thoughts here with a grain of salt.  I think it's better to do an XIT BS that makes the BS screen more relevant and has the appropriate links (e.g. PROD or CONSTRUCT) in a row there versus adding all of those links to BURN.  If someone took the XIT STO inventory bars (themselves from XIT FLT) and added them to XIT BS, you'd have a really potent buffer.  

My defense of ACT remains that its core to the workflow of burn.  You see you are low on a base, so you click act to address it.  

If one is adding shortcuts for base actions to BURN, I would draw the line where it currently is.

Thecat8413:
> I'm guilty of doing the added button to XIT BURN, so take my thoughts here with a grain of salt. I think it's better to do an XIT BS that makes the BS screen more relevant and has the appropriate links (e.g. PROD or CONSTRUCT) in a row there versus adding all of those links to BURN. If someone took the XIT STO inventory bars (themselves from XIT FLT) and added them to XIT BS, you'd have a really potent buffer.
> 
> My defense of ACT remains that its core to the workflow of burn. You see you are low on a base, so you click act to address it.
> 
> If one is adding shortcuts for base actions to BURN, I would draw the line where it currently is.

@jackinabox86 I don't think a XIT BS buffer would actually be that helpful over just using BURN. 

I would argue that most of the Buttons/Info in the BS buffer are rarely used. Most bases are built out all at once, and even when built piecemeal, it's still uncommon, so CONSTRUCT isn't needed. BUILDINGS only gets used every 60 days (or not at all, with XIT REP/ and BRA), WORKFORCE basically never, and EXPERTS is adjusted when you get a notification that a new one arrived or during base startup. The Workforce Overview, Developed/avail/total, and buildings icon list are essentially ignored. The only buttons on that menu that most people regularly use are PROD and INV. 

For most people, as far as bases go, the only things that are frequently used/opened are PROD, INV, and XIT BURN. Adding PROD to XIT BURN essentially removes the need for the native BS buffer in regular use. 

In that lens, your ACT button is actually pretty useful; for players who use XIT ACT, such a button would be a frequent flyer. 

If needed, adding both PROD and ACT buttons via a default-off setting might work, but I would expect most users not to know about the feature if not explicitly told it exists.

agm-114:
> Clicking the name of the base (when in default mode) currently expands the base info for XIT BURN. That's core functionality, so changing that doesn't really work in my mind.

The + l - button can be used to control size.

> Removing INV has the same issue as not adding PROD; it's a common action that a player would want easy access to. If anything, most users likely would want INV over PROD; most bases are "set-and-forget" for production, but inventory management is always needed.

If you are a basic / free user most bases are not set it and forget it.
Xit inv & prod are much better at statusing the underlying inventory issue.
My logic is since this is a prod menu, giving direct access to prod makes sense. Giving indirect access to all other buttons via a link that uses existing space seems like a good option. 
This is a prod buffer so having access to a prod makes some sense.
Just having inv enabled by default cause it's a common button, less so. If we do it will be enabled in all xit buffers that list bases, IE xit burn, xit prod, xit inc, xit full. If a user wants a list of inv buttons that is displayed without configuration they will be able to use xit inv.

Thecat8413:
> > Clicking the name of the base (when in default mode) currently expands the base info for XIT BURN. That's core functionality, so changing that doesn't really work in my mind.
> 
> The + l - button can be used to control size.

It can, and I suppose that is an option, but personally I find the UI to be better as is. In my experience, a small button is more frustrating than clicking a line for an end user. 

> 
> > Removing INV has the same issue as not adding PROD; it's a common action that a player would want easy access to. If anything, most users likely would want INV over PROD; most bases are "set-and-forget" for production, but inventory management is always needed.
> 
> If you are a basic / free user most bases are not set it and forget it. Xit inv & prod are much better at statusing the underlying inventory issue. My logic is since this is a prod menu, giving direct access to prod makes sense. Giving indirect access to all other buttons via a link that uses existing space seems like a good option. This is a prod buffer so having access to a prod makes some sense. Just having inv enabled by default cause it's a common button, less so. If we do it will be enabled in all xit buffers that list bases, IE xit burn, xit prod, xit inc, xit full. If a user wants a list of inv buttons that is displayed without configuration they will be able to use xit inv.

I would agree that XIT BURN is a prod menu, so having the PROD button does make a lot of sense. But BURN is also for monitoring consumables, so being able to move supplies around also makes sense to me. 

XIT PROD (at least the version in the live extension) isn't very helpful to me, to be honest. It gives a good overview in a compact form, but having the info in such a compact, tree heavy format is more annoying than anything else. I've tried using it and have honestly just switched back to using the default native PROD buffer. With one click, I can see the entire production line up, what is queued, and what is in progress without having to individually click on 'PP1', 'PP2','WEL', etc. 

PROD INV isn't currently live, so I haven't tested it at all yet. I do like the concept though, and I do have a INV buffer on my main page for accessing warehouses, so maybe it could work? I don't know. 

At the end of the day, I personally think leaving the INV button on is a good idea. Moving the base button to the base name is something I personally disagree with, but its also not necessarily a bad idea. 

Sorry for the ramble, I know this PR is getting a bit.... I don't even know what it call it.

PR #212, opened 2026-05-11
XIT ACT: Add Paste material group

erendrake:
## Summary

Adds a `Paste` material group that accepts pasted spreadsheet rows and parses them into a material bill — and, optionally, per-ticker price limits — at execution time. Useful for ad-hoc buy/ship lists copied from a spreadsheet without authoring a fixed `Manual` group.

The group has no design-time fields — materials are pasted in the configure dialog when the action package runs. To support that, `editComponent` is now optional on `MaterialGroupInfo` and `EditMaterialGroup.vue` skips rendering the edit form when none is registered.

## Paste format

Each row is `TICKER`, `QUANTITY`, and an optional `PRICE`:

```
RAT  100  530
DW   50
```

- **TICKER** — must be a known material.
- **QUANTITY** — positive whole number (a count of units). Grouping separators are stripped where unambiguous; a value that resolves to a non-integer (e.g. a stray decimal) is rejected.
- **PRICE** — optional, positive, **at most 3 significant figures**. The game rounds CX order prices to 3 sig figs (e.g. `43,810` becomes `43,800`), so finer prices are rejected with a suggested rounded value rather than silently truncated — what you paste matches the order that gets placed.

## Delimiters: TSV, CSV, or semicolon — one per paste, never mixed

Spreadsheet copies land on the clipboard as **TSV** (tab-separated), which is the unambiguous happy path. We also accept comma- and semicolon-separated text:

- **TSV (tab)** — the format Excel/Google Sheets produce when you copy cells. Decimals may use `.` or `,` (the tab is the column boundary, so there's no ambiguity).
- **CSV (comma, RFC 4180)** — a comma is always a column separator. A field that itself contains a comma **must be quoted**, e.g. `FLX,100,"35,4"`. This keeps `PT,7,529.94` unambiguously three columns.
- **Semicolon** — for European spreadsheets that export `;`-separated CSV with comma decimals (`FLX;100;35,4`).

The delimiter is detected once for the whole paste; **mixing delimiters is rejected** with a clear message. This avoids guessing locale from the text — which is impossible for bare comma input (`1,234.5` vs `1.234,5`), and moot anyway since neither is a valid 3-sig-fig price.

> [!NOTE]
> A single cell containing `FLX,100,35,4` (string-concatenated rather than separate cells) is **not** supported — it parses as four comma fields and is rejected. Copy separate cells (TSV) or quote comma-containing fields per RFC 4180.

## Prices → CX Buy price limits

When the paste includes a `PRICE` column, those prices are bridged into the action's per-ticker price limits. The material-group contract gains an optional `setPrices` side channel; the runner collects per-group prices and exposes them to actions via `getMaterialGroupPrices`. Prices are plain per-unit numbers — the currency is whatever exchange the CX action targets, so the paste group has no currency field of its own.

**Conflict guard:** if prices are set in **both** the paste group **and** the CX Buy action's manual price limits, the run hard-fails at execution with a message explaining the conflict and how to resolve it (keep prices in one place). This is enforced at execution because the pasted prices only exist at run time — a design-time check can't see them.

> CX Sell (#200) is not wired to consume paste prices in this PR; it needs the same bridge + conflict guard when it merges.

## Error feedback

The configure dialog validates live and reports **every** problem line, not just the first:

- A summary header — `N of M lines have errors`, or `X materials ready` when valid.
- A per-line list with the line number up front (`Line 3: …`), in a height-capped scrollable box so a large bad paste doesn't blow up the dialog.
- Mixed-delimiter pastes show a single banner instead of a per-line list.

## Test plan

- [ ] Create an action package with a `Paste` group + `CX Buy` action targeting it
- [ ] Run the package — confirm the configure dialog accepts pasted spreadsheet content (TSV)
- [ ] Confirm comma and semicolon pastes work; mixing delimiters is rejected
- [ ] Confirm validation rejects unknown tickers, non-integer quantities, malformed rows, and prices with >3 significant figures — listing all bad lines with line numbers
- [ ] Paste a `PRICE` column and confirm the CX Buy uses it as the price limit
- [ ] Set prices in both the paste and the CX Buy price limits — confirm the run fails with the conflict error


razenpok:
Hey @erendrake, I've rewritten the code to use PapaParse. I did some in-game checking, and it works as expected. I would appreciate it if you could do some testing on your side as well. Cheers!

PR #213, opened 2026-05-11
XIT ACT: Add shortDescription field to action and material group registries

erendrake:
## Summary

Adds an optional `shortDescription` field to `ActionInfo` and `MaterialGroupInfo` in the ACT registry, surfaced as a one-line description banner in the Edit Action and Edit Material Group dialogs.

The Type dropdown shows the action/group name only — when a package has multiple actions or groups of different types, you can't tell what each one does from the dropdown alone without selecting it and remembering. The description banner makes the choice self-explanatory.

Includes descriptions for all existing actions (`CX Buy`, `MTRA`, `Refuel`) and material groups (`Manual`, `Repair`, `Resupply`).

## Test plan

- [ ] Open an action package and add a new action — confirm description appears below the type dropdown
- [ ] Switch the type dropdown — confirm description updates
- [ ] Same for the material group editor


erendrake:
@razenpok I would love to have your thoughts on this one

razenpok:
Yep, looks good. I'll merge it in the next release

PR #216, opened 2026-05-16
feat(SHPI): add base inv and warehouse context bar buttons

jackinabox86:
Adds two context bar buttons to the SHPI buffer:

  - INV {planetId} — appears when the ship is landed at a planet where the player has a base; opens the base inventory
  - WAR {planetId} — appears when the ship is landed at a planet where the player has a warehouse; opens the warehouse inventory

  Both buttons are hidden while the ship is in flight or at a station.

  Shift-click behaviour

  Shift-clicking either button opens the target inventory as a horizontal companion buffer instead of a new floating buffer:
  - Floating tile: tile is widened by 450 px and split horizontally; the inventory command is set in the new right-hand tile
  - Already in a split: the sibling tile is reused and its command is replaced
  - Docked tile: no split occurs (same limitation as other companion-buffer features)
  - mousedown default is suppressed on shift to prevent browser text selection
   
https://github.com/user-attachments/assets/0ddee12b-c366-4fef-845f-25dd87eb2a55

### DIFF
--- a/src/features/basic/shpi-base-inv-button.tsx
+++ b/src/features/basic/shpi-base-inv-button.tsx
@@ -46,6 +46,9 @@ async function onTileReady(tile: PrunTile) {
     return (
       <div
         class={[C.ContextControls.item, C.fonts.fontRegular, C.type.typeSmall]}
+        onMousedown={(e: MouseEvent) => {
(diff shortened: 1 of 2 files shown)


razenpok:
Merged in 4714d6bf without the shift-click functionality

PR #217, opened 2026-05-21
Fix inconsistent maths in BURN Need column

richjenks:
Need column for planets used floored number of days which should only be for display so quantity was exaggerated. Fixed by using exact days so planet and overall rows match. Also added Need column to COPY now it's calculated reliably.


razenpok:
I've included your changes in 7966b823

PR #219, opened 2026-05-26
Fix showBuffer by improving lifecycle management for websocket

purpleorangeazure:
When opening a buffer, showBuffer checks if it can dispatchClientMessage to the currently active websocket by checking if the method is defined. Currently, dispatchClientMessage is defined as soon as a websocket is instantiated--before it has opened and messages can actually be dispatched. Because showBuffer does not allow concurrent invocations, this can lead to a situation where rPrUn waits indefinitely for a buffer that was requested during a period when a new websocket has been instantiated but not opened and can therefore not process the UI_TILE_CHANGE message.

This issue can occur when the active websocket is disconnected, a new one is created, and showBuffer is called in the period of time between the closing of the old websocket and the opening of the new one. During this period, dispatchClientMessage is defined but ineffectual.

This PR fixes this issue by setting dispatchClientMessage only when the active websocket is opened and back to undefined when the active websocket is closed, correctly allowing for showBuffer to use its fallback. 

One potential side effect of this fix is that dispatchClientMessage now calls the proxied onmessage event rather than the original one, but I have not observed nor do I foresee any issues created by this. 

Additionally, there is a chance that a new websocket is opened before the previous one has closed, leading to a situation where the closing of the old websocket clears the dispatchClientMessage of the new one. I am not familiar enough with websocket lifecycle management or the source code of PrUn to determine whether this is worth writing a check for.

### DIFF
--- a/src/infrastructure/prun-ui/buffers.ts
+++ b/src/infrastructure/prun-ui/buffers.ts
@@ -42,8 +42,9 @@ export async function showBuffer(command: string, options?: ShowBufferOptions) {
     }
   }
   await acquireSlot();
+  console.log('showBuffer', command, 'creating!');
   const create = await $(document.documentElement, C.Dock.create);
-
+  console.log('showBuffer', command, 'created!');
   try {
     const windows = document.getElementsByClassName(C.Window.window);
     const seenWindows = new Set(Array.from(windows));
@@ -59,8 +60,13 @@ export async function showBuffer(command: string, options?: ShowBufferOptions) {
       });
       create.click();
     });
+    console.log('showBuffer', command, 'got new windows!');
     await processWindow(newWindow, command, options);
+    console.log('showBuffer', command, 'processed window!');
     return newWindow;
+  } catch (e) {
+    console.error(e);
+    throw e;
   } finally {
     releaseSlot();
   }
@@ -84,12 +90,15 @@ function releaseSlot() {
 }
 
 async function processWindow(window: HTMLDivElement, command: string, options?: ShowBufferOptions) {
+  console.log('showBuffer', 'processWindow', command, 'Processing!');
   const input = _$(window, C.PanelSelector.input) as HTMLInputElement;
   const form = input.form;
   if (!form?.isConnected) {
+    console.log('showBuffer', 'processWindow', command, 'form not connected!');
     return;
   }
   if (!(options?.autoSubmit ?? true)) {
+    console.log('showBuffer', 'processWindow', command, 'not autosubmitting!');
     changeInputValue(input, command);
     return;
   }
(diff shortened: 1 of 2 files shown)


purpleorangeazure:
Also, I've noticed that the fallback for showBuffer can steal focus from the user's input while it is in use, such as when trying to MTRA while a XIT BURN is opening a bunch of BS buffers. It would be ideal if the fallback did not do this but it is a separate issue which should be very minor with normal internet latency. 
For reference on this entire PR, I am in Japan and there seems to be about a 750ms delay for the websocket to open under normal conditions, which is probably pretty close to the worst case for the issue.

purpleorangeazure:
This is evidently not the only mode of the issue, as it has recurred under unclear circumstances. I'll investigate it more later

purpleorangeazure:
Replicating my journal here from discord:
I am 90% confident in saying that the root cause of the issue lies in deserialize-prun-app.ts, which re-initializes the APEX script(s) at first load. This results in the creation of two simultaneous websockets. Seemingly, one of them works for dispatchClientMessage and one of them does not. I tried delaying the creation of the second websocket so that socket-io-middleware.ts could consistently grab the second one, but this did not seem to work and caused a flakier connection besides. Ideally, the first instance of the script wouldn't be allowed to run at all, but i'm not sure that's possible.
The issue seems to go away entirely when i comment out deserialize-prun-app.ts (successfully resumed from suspend eight times) 
Anyway, my best fix for this is an automatic retry on processWindow that always uses the manual input fallback
Also, the websocket that does not work will not work for anything, so it is possibly better to set dispatchClientMessage to undefined whenever it first fails
I definitely don't understand its purpose fully, but the fact that rPrUn causes two active websockets at a time has always felt wrong to me. My ideal solution for this would be to find a way to achieve the goal of deserialize-prun-app.ts without duplicating the running script. it feels like it might cause performance issues too, but I have no proof 
Also, surely two simultaneous active websocket connections per instance is way worse for the PrUn servers than an automated request here and there. Has this been approved?
Oh, and lastly, I set the fallback timeout to 100ms, which works for me and my PC. I did not test any lower timeouts, so it is possible that a lower number could be used or that this isn't long enough for a slower/more congested PC

src/infrastructure/prun-ui/buffers.ts
```diff
@@ -107,16 +106,27 @@ async function processWindow(window: HTMLDivElement, command: string, options?:
...
-  const selector = await $(window, C.Tile.selector);
+  let tileChanged = false;
+  setTimeout(() => {
+    if (tileChanged) {
+      return;
+    }
+    fallbackTileChange();
+  }, 100);
   await Promise.any([
     new Promise<void>(resolve => onNodeDisconnected(input, resolve)),
```
raylu, line 126: conflict here with a523a90d that https://mergiraf.org was able to solve

PR #222, opened 2026-06-17
Add Localization API

purpleorangeazure:
The lack of a localization API has driven, among other issues, a pattern of locating UI elements based on their index. This can be very fragile. Inspired by the simplicity of the CSS type, I set out to create a similarly easy-to-use way to interface with the game's localization. That resulted in this fork, which is mostly code generation. I had to fight typescript a lot more than I would have liked to get my ideal access pattern--where leaves and branches of a tree exist on the same level--but I believe that this is in a quite usable state. I have gone through each prior access of the i18N object to convert to the new method, and everything works as expected and is much more readable.


razenpok:
Good stuff, I was going to do that myself at one point, but your version is fine to iterate upon. Anything else you would like to do with it before changing this PR from draft to open?

purpleorangeazure:
Mostly, I would like someone with more experience in typescript to look over the LocalizationTree / Leaf types and make sure they're the best they can be. I should have written this into the PR but I feel like it is wrong that I have to coerce LocalizationTree to PrunLocalization and vice versa via unknown.

razenpok:
Ah, okay, then you can mark this PR as ready for a review - I will check and fix the TS issues if any

src/infrastructure/prun-ui/localization-type-generator.ts
```diff
@@ -0,0 +1,165 @@
...
+function emitLocalizationTree(
+  tree: LocalizationTree,
+  indent: number = 0,
+): `LiteralLocalizationLeaf` | `{${string}}` {
+  let result: string = ``;
+  const isTreeLeaf = isLeaf(tree);
+  const children = Object.entries(tree).filter(([key]) => !LEAF_KEYS.some(x => x == key));
+  const format = isTreeLeaf ? (tree as LocalizationLeaf).imf : undefined;
+  const formatOptions = format ? emitFormatOptions(format.getAst()) : undefined;
+  if (isTreeLeaf && format && formatOptions == 'void') {
```
purpleorangeazure: I'm 80% sure this needs a check for no children

razenpok:
So far so good! I haven't checked the i18n and localization-type-generator files yet, but the L convention and its usage in the features is solid, I like it 👍

purpleorangeazure:
Thanks! Looking at the replace function you've added to flight status icons, I'm wondering if there should be a helper in the vein of applyCssRule like patchLocalization(localization: LiteralLocalizationLeaf, patch: (string) => string) that can centralize that code

purpleorangeazure:
Also, I was considering better keys for LocalizationLeaves because imf is neither eminently readable nor obvious to look for if I want the format of the message. I think that it would be best to change to { getFormat(): () => IntlMessageFormat; message(...): () => string } so that getFormat, as a method, is obviously separate from sibling localization keys. I would use just format(), but I feel like that might look strange in the event you want to use the original format function, like .format().format() where both formats() are doing very different things. Also, message() is now clearly separated from imf.format(), which I think is good because it has a much simpler type signature so both should co-exist for different purposes.
I will submit a commit with the above changes

src/infrastructure/prun-ui/localization-type-generator.ts
purpleorangeazure: It did! This had caused some localization entries to be missing.

purpleorangeazure:
I added applyLocalizationPatch to unimport because applyCssRule is in there and it is of parallel function, but I wonder if this is necessary since localization patches happen much less often than new css rules. It's in a separate commit so it can be easily reverted.

razenpok:
patchLocalization is also a good addition to this stack. It's okay to include it in unimport, as my end goal is to properly support localization throughout the extension (both adding new strings and modifying the base game's ones), so it will likely be used in more places down the line.

What do you think about changing the leaf to be a function, instead of having a .message() function? Like this:
```js
// old
label?.textContent === L.MaterialInformation.ticker.message()

// new
label?.textContent === L.MaterialInformation.ticker()
```

src/infrastructure/prun-ui/i18n.ts
```diff
@@ -20,9 +94,83 @@ export function loadPrunI18N() {
...
+    for (const leafKey of LEAF_KEYS) {
+      switch (leafKey) {
+        case 'getFormat':
+          (cursor as Partial<LocalizationLeaf>)['getFormat'] = () => messageFormat;
+          break;
+        case 'message':
+          (cursor as Partial<LocalizationLeaf>)['message'] = messageFormat.format;
+          break;
+      }
+    }
```
razenpok: Is this chunk just a convoluted version of this code?
```ts
    const leaf = cursor as LocalizationLeaf;
    leaf.getFormat = () => messageFormat;
    leaf.message = messageFormat.format;
```

purpleorangeazure: Yes, it is. I used a switch statement so that my IDE would warn me if I was using incorrect keys or I was missing one as I was frequently changing the LEAF_KEYS while I was working on this (hence them being in a top-level const). That's probably unnecessary complexity now, though.

purpleorangeazure:
I thought of doing that, but I am not used to functions not being end-points in the literal sense of the word, so I felt it would make it unclear when a localization key was both a leaf and had children. I tried implementing it just now, and I think that the added concision actually makes it more clear what exactly the localization id is. Also, option type annotations are slightly more clear since they've been promoted to the very first part of the type. I will commit as soon as servers are back up from maintenance and I can check that everything still works.

purpleorangeazure:
I ran into an issue with assigning a function to an object--this does not make the object callable, so I had to change the generateLocalizationTree logic a bit. I got rid of LiteralLocalizationLeaf in the generated code because it was obscuring clear function signatures. Also, I noticed that the intellisense on leaves with children gets very cluttered:
image

Old version:
image


I think that an explicit message function may strike a better balance between child visibility and call clarity after all, but it's up to you.

razenpok:
I've reworked the types generator to use existing types and use really short type aliases to save on the file size (500kb => 200kb). The ide still shows pretty type names, so it works fine with the short ones.
image

I've also split the localization types and template generators. The reasoning here is that your compiler always needs types, whereas you need templates only occasionally, when looking for that one specific string. So, to save on the file size here as well, we can just generate the templates separately. I've added the generator buttons to XIT DEV for that.

> I think that an explicit message function may strike a better balance between child visibility and call clarity after all, but it's up to you.

In my opinion, in the case of this L object, you don't really need intellisense, nor member discoverability, nor child visibility. All because the main development flow will be: you see a localized string => you search for the localization key  in the templates file => you use the key to access L and write localization-related code => you never touch this code ever again, probably. IntelliSense is a good thing for programming APIs, where you discover (or rediscover) the API surface. Here, you will probably never go looking "hmmm I wonder which members are there in L.WhateverPanel.table". The goal of the L object is to provide typed access to the localization tree.

razenpok:
And now accessing a missing localization key won't throw in runtime %)

razenpok:
I will probably do a final review and merge tomorrow, so let me know if you want to add or change smth before I do that

purpleorangeazure:
I looked over the code and three things stood out to me:
1. IsLeaf should probably just be checking if the object is a function now
2. applyLocalizationPatch should probably be entirely replacing the AST to handle strings that have tags like "this is a \<link>string\</link>" rather than just single-element literals, lest it result in confusing or undesired behavior.
3. LEAF_KEYS probably isn't necessary and should be removed for simplicity.

Other than that all looks good to me, thank you for your work integrating this PR!

(I will commit each of these later)

purpleorangeazure:
I ended up removing isLeaf and LEAF_KEYS as the former was used only once in the type generator (even though it was declared in the main file) so it was better to inline and the latter was an array of one element that wasn't very likely to change from now on.

razenpok:
Merged, thank you!

PR #224, opened 2026-06-23
Multilingual localization patch

purpleorangeazure:
A simple change to localization patching to better support situations like bs-hide-zero-workforce where it was previously necessary to find and replace the english text so as to not mess with other languages. I added paths for en, es, and de since those were the languages I felt I could confidently work with.
I made sure to keep the behavior of not touching the AST of unpatched languages at all unless a default is specified.


purpleorangeazure:
I also had the idea of applyLocalizationPatch taking a type argument of the localization's parameters like `applyLocalizationPatch<T>(localization: ParametrizedLocalizationLeaf<T>, patch: (options: T) => string)` so that it would not be limited to LiteralLocalizationLeaves but I decided against implementing it because I don't know of anywhere that would be used.

purpleorangeazure:
I have finished transferring what is possible of lm-clean-ads over and I hope that demonstrates the immense readability and flexibility benefits of this PR

razenpok:
Yep, looks exactly like what I was looking for. I will add it in the next release, though - I'm preparing something big for the current one.

PR #225, opened 2026-06-24
XIT BURN: Fix TSV export reporting double inventory

erendrake:
The TSV export computed inv as mat.inventory + mat.inventory, doubling the reported inventory. Use mat.inventory + mat.remainingAllocation to match the displayed inventory and the daysLeft calculation.

Proposed changelog entry
### Fixed

- `XIT BURN`: Fixed the TSV export reporting double the inventory in the "Inv" column


razenpok:
Whoops, copy-paste error. Thank you for the fix!

PR #226, opened 2026-07-02
XIT BURN: Fix color filter mismatch and inventory +1 rounding

### LINKED ISSUE #228: XIT burn fractional days confuse color selector
### Description

Within `xit burn` the selector does not show things in yellow state with fractional dates greater than than the setting.

Burn Settings:

* Red : 4
* Yellow : 8
* Resupply : 14

Image

Goods that have a burn in the range (8.0, 9.0) are colored yellow.

Example QQ-001b: 

Image

Image

I expected to see sodium (5.9), hydrogen (8.3) and boron (8.8) in the list when green was unselected.  I only saw sodium (5.9).

Example XH-594a

Image

Image

I expected to see steel (8.5) in the list.

This bug also extends into red settings. Reducing the supply of steel to 4.4 days of supply it is colored red but shows up in the yellow selector.

Image

---

The range for yellow set at 8 and red at 4 should be (4.0, 9.0] rather than (4.0, 8.0).   
The range for red set at 4 should be [0.0, 5.0) rather than [0.0, 4.0).

1. Reduce a good's amount a fractional amount more than the number of days the burn color (yellow or red) is set for.
2. Remove the green (or yellow) from the filter.
3. Observe that the good with a fractional day more than the configured value does not show up in the expected filter.

### Extension version

26.6.22.2033

### Browser(s) used

150.0.7871.115 (Official Build) (arm64) ; OS: macOS Version 26.5.2 (Build 25F84) ; Javascript V8 15.0.245.15

erendrake:
Fixes two XIT BURN display bugs that share a root cause: `7966b823` ("fractional allocation support") pushed fractional precision into the days/inventory math for accuracy, but the display in two places drifted so it **overstates** what you actually have.

## 1. Color filter didn't match the cell color

The RED/YELLOW/GREEN filter buttons classified rows using **raw** days while the cell color **floors** them. A 7.1-day row displayed as yellow (floor 7 ≤ yellow threshold 7) but was filtered as green, so unchecking YELLOW hid a row that looked red/yellow.

Fix: extract `getBurnThresholds()` as the single source of truth and use it in both `DaysCell` (color) and `MaterialRow` (filter) so they can't drift again. Classification compares true days against the thresholds, mutually exclusive: red `days ≤ R`, yellow `R < days ≤ Y`, green `days > Y`. This intentionally replaces the old floor-based coloring (1.7 with red threshold 1 was red, now yellow; 8.3 with yellow threshold 8 was yellow, now green) and drops the cumulative red-implies-yellow filter behavior.

## 2. Inventory column rounded up past the game amount

The `Inv` column formatted fractional inventory (`inventory + remainingAllocation`) with `Intl` formatters that **round**, so a raw 19.95 showed as **20** while the game floors partial stacks to **19** — the reported "+1" (e.g. HMS 20 vs 19, COF 61 vs 60).

Rounding up also overstates runway, which is the exact regression fractional-allocation support was meant to fix: previously a real 0.15-unit stack could show as 1 and imply days of burn you didn't have, so you'd run out earlier than BURN predicted.

Fix: truncate to the displayed precision instead of rounding, so the shown amount never exceeds what you actually hold. `daysLeft` still uses full precision internally, so runway accuracy is unchanged; only the display stops rounding up.

## Verification

- `tsc --noEmit`, eslint, prettier all clean.
- Threshold cases (red 1, yellow 8): 0.9 → red, 1.0 → red, 1.7 → yellow, 8.0 → yellow, 8.05 → green; 10.9 → green with a threshold of 10.
- Inventory cases checked: 19.95 → "19.9" (was "20"), 60.95 → "60.9" (was "61"), 0.15 → "0.15", 109.6 → "109"; previously-correct values unchanged.


d-bucur:
Tried this locally and I can confirm it solves point 1. These now display on red, whereas before they didn't:

Screenshot from 2026-07-04 23-13-59

I still find it odd though that with a red threshold of 1, 1.7 and 2 still display as red. In my mind 1.7 should be yellow and 2 should most definitely not be red. Seems like a rounding error somewhere.

Also I don't agree with the cumulative behavior, ie isRed => isYellow. These conditions should be mutually exclusive; that is why the cells have only a single color, that is why the filters are separate and independent. By this logic, isYellow should also imply isGreen for consistency, which would defeat the entire point of separate filters.

The way I would expect the thresholds to work is:
- red: x <= R
- yellow: R < x <= Y
- green: x > Y

erendrake:
@d-bucur Thanks for testing it, and you were right on both counts.

I've pushed an update that drops the cumulative behavior and implements exactly the thresholds you describe, mutually exclusive: red days ≤ R, yellow R < days ≤ Y, green days > Y. The earlier version also floored days to whole numbers before comparing, which is why 1.7 with a red threshold of 1 still read as red — that's gone too. 1.7 now lands in yellow, and 2 is red only if you've set red to 2 or higher.

Cell color and filter buttons both go through the same classification, so they can't drift apart again. PR description updated to match. This also lines up with what the reporter hit in #228.

Issue #234, opened 2026-08-05
Add option for suppressing "Name" in INV


shagie:
### Description

For most users, the 'Name' column in the INV panel appears to be unused and contains the value '--' for everything.

<img width="628" height="297" alt="Image" src="https://github.com/user-attachments/assets/1661b221-4bca-49b4-a1e7-193e9dd09555" />

Add an option to the settings that would suppress this column similar to the functionality of `flt-hide-transponder` allowing a user to reclaim some horizontal space.

razenpok:
PR #237 added ship names to the Name column

PR #237, opened 2026-08-13
Add inv-ship-destination-location and inv-ship-names

lumivient:
1. Shows ship names and locations in INV
2. Shows destination when in-flight (e.g. `→Boucher`)
3. Reacts to flight change/abort

Before:
before

After:
after


raylu:
#234 suggested the opposite
I think this is better, though

lumivient:
Agree, closer to expected behaviour anyway http://prosperousuniverse.com/blog/2025/04/28/smoothing-edges-481

PR #238, opened 2026-08-26
Add XIT FLOW

erendrake:
New `XIT FLOW` buffer: one row per material across your bases — daily production, consumption (incl. workforce), net delta, its value per day, and which bases produce/consume it. `XIT FINPR` shows per-base profit; this shows *which goods* the money flows through.

image

Rates come from `core/burn.ts` unchanged, so Prod/Cons match `XIT BURN OVERALL` exactly.

## Value

Net deficit × Buy (Ask), net surplus × Sell (Bid). MM materials sell at the MM bid and buy at the cheaper of the Ask and the MM ask. Ignored materials value at 0; a ticker with no market data falls back to the regular Refined PrUn price (0 when absent from the CX), so `--` only appears before price data loads; those rows sort last in the Value sort.

image

## Overrides

Click a Buy or Sell price to type your own (partner deals). Enter/blur saves, Esc cancels, empty clears. Marked `*`; stored in `settings.flow.overrides`; **FLOW-only** — FINPR/FINBS/MAT are untouched. Invalid input (non-numeric or negative) keeps the previous value; a plain text input rather than `NumberInput` because empty must mean "clear".

image

## Also

- `XIT FLOW <planet…>` / `XIT FLOW NOT <planet…>` — BURN's parameter grammar.
- Click headers to sort (Ticker uses BURN's category order); persisted per tile. Sticky headers.
- COPY → TSV.

image

image

## Review notes

- Outside the feature: `getMarketPrices` (Ask/Bid pair, same MM/ignored rules as `getPrice`) in `cx.ts`, `settings.flow.overrides` + one migration. No existing behaviour changes.
- Sticky header uses the same `#222222` as `XIT SET FEAT`.
- Verified in-game: Prod/Cons = BURN OVERALL; ΣValue (before the buy/sell split) = FINPR Profit + Repairs.
- Commits: buffer → settings/getters → pricing with overrides → review feedback. `CHANGELOG.md` untouched.


erendrake:
@razenpok please let me know what you think

razenpok:
Looks cool! Will probably add it in one of the next releases

PR #239, opened 2026-09-03
Add XIT TOTALS command

EthanP1337:
Adds `XIT TOTALS`, a panel that shows how much of each material you hold
across your whole company, with optional per-exchange valuation.

`XIT BURN` answers "how long until I run out on this planet". This answers
"how much of this do I own anywhere, and what is it worth". Today that means
adding up base inventories, warehouses and ship holds by hand.

## Usage

- `XIT TOTALS` — the preset list. Create, rename, reorder and delete presets.
- `XIT TOTALS <name>` — show a preset.
- `XIT TOTALS EDIT <name>` — edit the ticker list of a preset.

Presets are stored in `userData.itemTotalsPresets`.

image
image

## Columns

`Item | Amt | Weight | Volume | <one per enabled exchange>`

- **Amt** sums every store the company owns, as listed above.
  `storagesStore` already excludes construction and upkeep stores.
- **Weight / Volume** are `amount × material.weight` and
  `amount × material.volume`, formatted as `t` and `m³` to match
  `MaterialPurchaseTable`.
- **Exchange columns** show what the whole holding is worth at that exchange,
  in that exchange's own currency, for example `12,400 CIS`. The currency is
  in the cell rather than the header, because the six exchanges use four
  different currencies and the numbers are not comparable across columns.

## Toggles

Toggles live in a `C.ComExOrdersPanel.filter` bar and persist per tile
through `createTileStateHook`, the same as `XIT BURN`.

- **W/V** — adds the Weight and Volume columns.
- **AI1 / CI1 / CI2 / IC1 / NC1 / NC2** — one per exchange, each adds that
  exchange's value column. The list is built from `exchangesStore`, and the
  enabled codes are stored as a string array, so a new exchange would appear
  on its own with no code change.
- **TOTAL** — adds a summary row using the game's own
  `C.IncomeStatementPanel.totals` divider, as `XIT FINBS` does.

The TOTAL row deliberately leaves **Amt** blank. Adding units of different
materials together gives a number with no meaning. Weight, volume and the
currency values do share a unit, so those are summed. A value column totals
to `--` if any item in the preset has no price there, so a partial sum is
never presented as a real number.

## Shared code

`src/infrastructure/fio/cx.ts` gains `getPriceAt(ticker, exchangeCode)`.
The price data was already keyed by exchange; `getPrice` could only read the
one exchange configured in `XIT SET`. `getPrice` now delegates to
`getPriceAt` with that same configured exchange, so existing callers are
unchanged. Valuation honours the user's configured pricing method, ignored
materials and MM materials exactly as before.

## Relationship to #236 (XIT STOCK)

These are complementary, not competing. They answer different questions and
share no code.

**#236 is per-location.** You configure planets, and per planet a list of
tickers with a target amount. It sums that planet's base inventory, plus that
one location's warehouse when the per-item toggle is on. It then reports
current/max, a percentage, and days-to-target from the burn rate. The
question it answers is "am I stocked up on Katoa".

**This PR is company-wide.** One row per material, summed across *every*
source the company owns:

- every planetary base inventory
- every rented warehouse, on planets and on CX stations
- every ship cargo hold
- every ship fuel tank

There are no targets and no per-location breakdown. The question it answers
is "how much of this do I own in total, and what is it worth". It adds
per-exchange valuation, which #236 does not have.

Both are multi-preset XIT screens and both follow the routing convention of
`XIT ACT`.

**Merge note:** if both land, whichever merges second needs a one-line
rebase. Both add a field after `actionPackages` in
`src/store/user-data.ts` and an interface block in
`src/store/user-data.types.d.ts`. Nothing else overlaps.

## Notes

- No `CHANGELOG.md` change, per `docs/contributing.md`.
- No new dependencies.
- Verified: `tsc --noEmit`, `eslint` and `vite build` all clean.

### DIFF
--- a/src/core/item-totals.ts
+++ b/src/core/item-totals.ts
@@ -1,0 +1,73 @@
+import { storagesStore } from '@src/infrastructure/prun-api/data/storage';
+import { materialsStore } from '@src/infrastructure/prun-api/data/materials';
+import { exchangesStore } from '@src/infrastructure/prun-api/data/exchanges';
+import { getPriceAt } from '@src/infrastructure/fio/cx';
+
+export interface ItemExchangeValue {
+  code: string;
+  name: string;
+  currency: string;
+  // Worth of the whole holding at this exchange. Undefined when there is no price.
+  value?: number;
+}
+
+export interface ItemTotalStatus {
+  ticker: string;
+  material?: PrunApi.Material;
+  // Total amount across every inventory the company owns.
+  amount: number;
+  // Total weight in tonnes and total volume in cubic metres.
+  weight: number;
+  volume: number;
+  values: ItemExchangeValue[];
+}
+
+// Every exchange, ordered by code, so the columns keep a stable order.
+export const exchanges = computed(() =>
+  [...(exchangesStore.all.value ?? [])].sort((a, b) => a.code.localeCompare(b.code)),
+);
+
+// Sums every material across all owned inventories: bases, warehouses,
+// ship cargo holds and fuel tanks. Construction and upkeep stores are
+// already excluded by storagesStore.
+const totals = computed(() => {
+  const stores = storagesStore.all.value;
+  if (!stores) {
+    return undefined;
+  }
+  const totals = new Map<string, number>();
+  for (const store of stores) {
+    for (const item of store.items) {
+      const quantity = item.quantity;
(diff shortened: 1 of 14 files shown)


EthanP1337:
I don't love the name "Totals" open to suggestions

PR #241, opened 2026-09-07
Add unclipped-tooltips

purpleorangeazure:
PrUn clips tooltips whenever they overflow the bounds of their tile:
image

What if it didn't:
image

Also handles screen edges:

image

image

This is achieved by hiding the regular tooltips and moving around, showing, and hiding a popover placed at the document root. It is fully compatible with all tooltips as it relies on the same data-tooltip attribute to manage.

Currently, this does not draw the tooltip arrows, but it definitely could. Nor did I remove the prun-bugs styles to fix those arrows with the unmodified method.


purpleorangeazure:
image
Came back and implemented the triangle, along with a more robust positioning algorithm. I kept the gap between the pointer and the tooltip, but I don't really like it. Were it up to me, I would remove the gap :P

razenpok:
Goated

razenpok:
I've extracted it into a separate feature. Too big for a "collection of small prun bugfixes" :D

purpleorangeazure:
``white-space: pre-wrap`` is causing the tooltip to look... odd in my test case of the "inv" tooltip of XIT BURN, at least.
image

Is this an acceptable tradeoff for keeping exact white space from the source?

In this specific case, it can be resolved by removing the explicit new line from BURN.vue.

razenpok:
Hmmm, indeed. It was a fix for a different issue: XIT PROD uses newlines, and they did not render without pre-wrap
image

purpleorangeazure:
Seems like the right way forward then is to remove superfluous new lines like the one in XIT BURN rather than remove pre-wrap.

razenpok:
Or we could just synchronize the value of `white-space` with the new tooltip

purpleorangeazure:
Frankly I'm not sure what you mean but whatever works ☺️

razenpok:
Try the latest commit in main

